PR #24, opened 2018-12-08
Integrate tree view

Divyata-Singh:
- Integrated treeview with boolean search
- Added a config file in treeview. The config file has an object for callback functions on node click
- Added a Readme in treeview for explaining how to use the node click callbacks

### DIFF
--- a/boolean-search/booleansearch.app.js
+++ b/boolean-search/booleansearch.app.js
@@ -1,3 +1,4 @@
+var setSourceForFilter;
 (function () {
     'use strict';
     class filterModel {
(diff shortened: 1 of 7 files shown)


Divyata-Singh:
@jrchudy Could you please review the changes I made in treeview. I have just added a callback on node click based on the Parent_App query parameter in the treeview url. The default node click behavior ( when Parent_App query parameter is absent) is still the same. My changes for treeview are hosted here: https://dev.rebuildingakidney.org/~divyatas/treeview/index.html and for boolean-search: https://dev.rebuildingakidney.org/~divyatas/boolean-search/index.html

PR #59, opened 2019-05-29
test cases for heatmap in deriva-webapps project

iamyle0710:
1. created general test cases which includes checking link content, link tooltip and heatmaps visibility.
2. created test cases for 4 heatmaps, which includes checking their title, legend visibility, legend values, number of sections
3. created README file of how to run E2E tests for Deriva-webapps (reference to RBK's README file)


iamyle0710:
@jrchudy PR updated

test/e2e/specs/heatmaps/00-heatmap.spec.js
```diff
@@ -0,0 +1,379 @@
...
+
+        it('legend should have values: `9.5`, `10`, `10.5`', function(){
+            heatmapElem
+            .first()
+            .all(by.css(".infolayer .crisp text"))
+            .getText()
+            .then(function(textArr){
+                var legendValues = []
+                for( var i = 0; i < textArr.length; i++){
+                (function(index){
```
jrchudy: You shouldn't need to define an anonymous function here. The for loop isn't creating any promises that would be executed later at a different time. If you are unsure of what I mean, I can go over it in more detail. 

There are some places in our existing tests where we loop have loops that surround `describe` blocks and other function calls that generate promises where capturing the value of the variable (index in most cases) is important because that function will be executed at some point later in time rather than right after the loop is called.

iamyle0710:
> The tests look good. I left one comment in the code about creating an anonymous function. Can you also change the folder name from `heatmaps` to `heatmap` so it's consistent with the appname and the folder at the top level of the `deriva-webapps` repo?
> 
> Another comment I have is that you could actually simplify this test file by writing a loop where in each iteration of the loop it runs the tests for 1 heatmap, and goes down the page testing the rest of the heatmaps in each of the following iterations. You don't have to make this change, but if you would like to you can.

iamyle0710:
@jrchudy Thanks for your help. I just updated to what you suggested.

PR #60, opened 2019-05-29
test(core): test cases for boolean search webapp

vipul-21:
test case for:
- Navbar display test cases
- Treeview loading test case
- Boolean Search: table header and default value of the content of the first row


vipul-21:
@jrchudy PR updated.

vipul-21:
@jrchudy Updated the PR with new folder structure.

PR #65, opened 2019-06-06
Treeview testcases

vipul-21:


test/e2e/specs/treeview/00-dynamic-treeview.spec.js
```diff
@@ -0,0 +1,254 @@
...
+          });
+
+          facet1.all(by.tagName('img')).getAttribute('src').then(function (linkArr) {
+            linkArr.forEach(function (url) {
+               expect(imgFullLinks).toContain(url);
+               httpFun.httpGet(url).then(function (res) {
+                     expect(
+                       res.statusCode).toEqual(200, url + " is not a valid image source.");
+                     done();
+                   });
```
jrchudy: In the places that you have added `done`, make sure to have a catch block that calls `done.fail()`. So in this case, it could look like:
```js
httpFun.httpGet(url).then(function (res) {
  expect(res.statusCode).toEqual(200, url + " is not a valid image source.");
  done();
}).catch(function (err) {
  console.log(err);
  done.fail();
});
```

This will make it easier to debug if there is an error in the `httpGet` function.

test/e2e/specs/treeview/00-record-page.spec.js
```diff
@@ -0,0 +1,50 @@
...
+    it('verify that treeview are displayed', function () {
+      if (ele.isElementPresent(by.tagName('iframe'))) {
+        iframe = ele.element(by.tagName('iframe'));
+        iframe.getAttribute('src').then(function (iframeUrl) {
+          httpFun.fireHttpRequest(iframeUrl).then(function (res) {
+            expect(res.statusCode).toEqual(200);
+          }).catch(function (err) {
+            console.log("Error in promise chain: ", err);
+          });
+        });
```
jrchudy: This should be changed to use promise chaining so that the `catch` block only needs to be defined once.
```js
iframe.getAttribute('src').then(function (iframeUrl) {
  return httpFun.fireHttpRequest(iframeUrl)
}).then(function (res) {
  expect(res.statusCode).toEqual(200);
}).catch(function (err) {
  console.log("Error in promise chain: ", err);
});
```

vipul-21:
I think you're right. We should use async/await like :
` await browser.get(baseUrl+ testConfig.treeviewPage.dynamic);`
It's your call, I can update the PR with the calls. Still trying to figure out how to replace it for every waiting elements we have.

Regarding inverse of the await, I think it behaves the same way in terms of normal await. If the promise is resolved, it will negate it i.e make it `false`. For example, 
```
function resolveAfter2Seconds(x) {
  return new Promise((resolve, reject) => {
    setTimeout(() => {
      resolve(x);
    }, 2000);
  });
}
async function foo() {
  try{
    var x = !await resolveAfter2Seconds(10);
    console.log('We got this resolved', x);
  } catch (e) {
    console.log('We have to reject this one', e);
  }
}
foo();
```
And in case of reject, we need to catch the error(in both !await and await).

jrchudy:
> And in case of reject, we need to catch the error(in both !await and await).

If we still need to catch the error, as long as those calls are done within a promise, it should be simple enough:
```js
element.getInnerHtml().then(function (html) {
  await someCondition();

  expect(html).toBe("some html");
}).catch(function (err) {
  ...
});
```

But in some cases (in beforeAll) those cases are outside of a promise chain so we would have to do something like:
```
try {
  await browser.get(baseUrl+ testConfig.treeviewPage.dynamic);
} catch (err) {
  ...
}
```

Although we aren't doing that now when an error is thrown because of a wait condition, that generally will cause protractor to respond with the right error. But it is something we should always try to do.

It would be nice if you could make those other changes in this PR to use `await`/`async` whenever possible instead of the old "wait conditions". That way when we start developing more tests going forward, we can rely on `await`/`async` with more confidence.

vipul-21:
Regarding the `async/await` changes suggested, there are few advices to keep in mind if we want to use it.
1) Make sure you disabled control flow / promise manager: https://github.com/angular/protractor/blob/master/lib/config.ts#L631 Mixing awaits with enabled control flow may lead to unpredictable results.

2) Do not forget to prepend ALL your async actions with await (usually this is all protractor api methods). If you will forgot to do this - action without await won't be queued with other actions, so order of actions will be broken.

Source: https://stackoverflow.com/questions/44691940/explain-about-async-await-in-protractor/44701633#44701633

Note: Current implementation only uses `async/await` for viewport.

vipul-21:
@jrchudy Have updated the chaise repo doc. And have raised the PR for the same.

vipul-21:
Deleting this branch as the current test cases are being done on `functional-testcases`.

PR #66, opened 2019-06-11
test($core): functional testcases for treeview

vipul-21:
Static treeview
- hierarcy testcase for opening and closing of node list
- action buttons: search and expand alland collapse all button test
- filter terms based on sepcific Theiler Stage

Dynamic treeview
- hierarchy testcase for opening and closing of node list
- search feature test
- error messages correctly displayed

### DIFF
--- a/test/e2e/specs/treeview/00-dynamic-treeview.spec.js
+++ b/test/e2e/specs/treeview/00-dynamic-treeview.spec.js
@@ -6,7 +6,7 @@ var expectedValue = require('./dynamic-treeview-expected-value.js');
 exports.tests = function (appName, baseUrl) {
 
   describe('Dynamic page '+ appName, function () {
-
+    var EC = protractor.ExpectedConditions;
     beforeAll(function () {
       browser.ignoreSynchronization=true;
       browser.get(baseUrl+ testConfig.treeviewPage.dynamic);
@@ -243,7 +243,21 @@ exports.tests = function (appName, baseUrl) {
             expect(pageTitle.isDisplayed()).toBe(false, 'Page title is displayed');
         });
 
-        describe('Actions for the page', function() {
+        it('it is displayed with no errors', function() {
+          expect(element(by.id('jstree')).isDisplayed()).toBeTruthy('Treeview is not visible.');
+          expect(element(by.id('error-message')).isDisplayed()).toBe(false);
+        });
+
+        it('it should auto scroll to the first annoted term of the tree  (nerve of detrusor muscle of bladder)', function() {
+          var ele = element(by.id('nid_1431'));
+          expect(deriva.elementInViewport(ele)).toBeTruthy('first annoted term of the tree is not visible.');
+        });
+
+
+        describe('actions for the page', function() {
+          beforeAll(function() {
+            searchInput = element(by.id('plugins4_q'));
+          });
 
           it('dropdown value should be TS26: 18 dpc by default', function() {
             expect(element(by.id('number-button')).getText()).toEqual('TS26: 18 dpc');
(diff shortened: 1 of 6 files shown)


test/e2e/specs/treeview/00-static-treeview.spec.js
```diff
@@ -37,23 +37,132 @@ exports.tests = function (appName, baseUrl) {
...
+
+        });
+
+        it('all node list collapse on `Collapse All` click', function(done) {
+          actionBtn.all(by.css('button')).then(function(buttons) {
+            buttons[1].click().then(function() {
+              expect(element(by.id('jstree')).all(by.css('.jstree-open')).count()).toBe(0);
+              done();
+            });
+          }).catch(function(err) {
```
jrchudy: Similar comment as last time, you should avoid nesting promises inside each other. This case will work fine (because of your use of done) but it's better to control the flow of the test case using promise chaining:
```js
actionBtn.all(by.css('button')).then(function(buttons) {
  return buttons[1].click();
}).then(function() {
    expect(element(by.id('jstree')).all(by.css('.jstree-open')).count()).toBe(0);
    done();
}).catch(function(err) {
```

test/e2e/specs/treeview/00-static-treeview.spec.js
```diff
@@ -37,23 +37,132 @@ exports.tests = function (appName, baseUrl) {
...
+          });
+        });
+
+
+        it('all node list expand on `Expand All` click', function(done) {
           actionBtn.all(by.css('button')).then(function(buttons) {
-            expect(buttons.length).toEqual(2, '2 buttons are not visible.');
-            expect((buttons[0]).getText()).toEqual('Expand All', 'Expand All button is not correct.');
-            expect((buttons[1]).getText()).toEqual('Collapse All', 'Collapse All button is not correct.');;
+            buttons[0].click().then(function() {
```
jrchudy: This test case fails for me most of the time with this message:
`Expected 0 to be 4.`

I'm guessing it's checking for the "sub-tree" elements before they have finished loading in the DOM. Try adding a wait for condition here. You could try 2 different things, the 1st should work, the 2nd definitely will:
1. Wait for `.jstree-open` node to be visible
    ```js
    browser.wait(EC.presenceOf(element(by.id('jstree')).first(by.css('.jstree-open'))))
    ```

2. Wait for all `.jstree-open` nodes to be visible
    ```js
     browser.wait(function() {
       return element(by.id('jstree')).all(by.css('.jstree-open').count().then(function(ct) {
         return (ct==4);
       });
    }, browser.params.defaultTimeout);
    ```

The same comment about promise chaining applies here too.

test/e2e/specs/treeview/00-static-treeview.spec.js
vipul-21: Okay, will keep that in mind and use promise chaining. Sorry for that.

test/e2e/specs/treeview/00-static-treeview.spec.js
```diff
@@ -37,23 +37,132 @@ exports.tests = function (appName, baseUrl) {
...
+              expect(element(by.id('jstree')).all(by.css('.jstree-open')).count()).toBe(4);
+              done();
+            });
           }).catch(function(err) {
             console.log(err);
```
jrchudy: make sure to include `done.fail()` as well

test/e2e/specs/treeview/00-static-treeview.spec.js
```diff
@@ -37,23 +37,132 @@ exports.tests = function (appName, baseUrl) {
...
+
+      });
+
+      it('expand all and collapse all button are visible', function(){
+        actionBtn.all(by.css('button')).then(function(buttons) {
+          expect(buttons.length).toEqual(2, '2 buttons are not visible.');
+          expect((buttons[0]).getText()).toEqual('Expand All', 'Expand All button is not correct.');
+          expect((buttons[1]).getText()).toEqual('Collapse All', 'Collapse All button is not correct.');;
+        }).catch(function(err) {
+          console.log(err);
```
jrchudy: make sure to include `done.fail()` as well

test/e2e/specs/treeview/00-static-treeview.spec.js
```diff
@@ -37,23 +37,132 @@ exports.tests = function (appName, baseUrl) {
...
+        });
+
+        it('all node list collapse on `Collapse All` click', function(done) {
+          actionBtn.all(by.css('button')).then(function(buttons) {
+            buttons[1].click().then(function() {
+              expect(element(by.id('jstree')).all(by.css('.jstree-open')).count()).toBe(0);
+              done();
+            });
+          }).catch(function(err) {
+            console.log(err);
```
jrchudy: make sure to include `done.fail()` as well

test/e2e/specs/treeview/00-static-treeview.spec.js
```diff
@@ -37,23 +37,132 @@ exports.tests = function (appName, baseUrl) {
...
+      });
+
+      describe('actions for the page', function() {
+        beforeAll(function() {
+          browser.wait(EC.elementToBeClickable(filterDropDown), browser.params.defaultTimeout);
+          filterDropDown.click();
+          browser.wait(EC.visibilityOf(element(by.id('number-menu'))), browser.params.defaultTimeout)
+          element(by.id('ui-id-2')).click();
+
+          deriva.waitForElementInverse(element.all(by.id("loadIcon")).get(0));
```
jrchudy: You should use chained promises for click events as well so that the control flow will work properly. Each `browser.wait()` and each `.click()` push a promise into the queue to be executed at a later time. The way you wrote this test won't cause any problems because each wait/click condition will be acted on in the same order they were pushed.

```js
browser.wait(EC.elementToBeClickable(filterDropDown), browser.params.defaultTimeout);
filterDropDown.click().then(function () {
  browser.wait(EC.visibilityOf(element(by.id('number-menu'))), browser.params.defaultTimeout)
  return element(by.id('ui-id-2')).click();
}).then(function () {
  deriva.waitForElementInverse(element.all(by.id("loadIcon")).get(0));
}).catch(...);
```

This chaining will ensure that the click events are triggered when we want them to be. If you want to be more explicit, then we can add the wait commands to the promise chain as well:
```js
browser.wait(EC.elementToBeClickable(filterDropDown), browser.params.defaultTimeout).then(function () {
  return filterDropDown.click()
}).then(function () {
  return browser.wait(EC.visibilityOf(element(by.id('number-menu'))), browser.params.defaultTimeout);
}).then(function () {
  return element(by.id('ui-id-2')).click();
}).then(function () {
  deriva.waitForElementInverse(element.all(by.id("loadIcon")).get(0));
}).catch(...);
```
This change doesn't need to be done, but it uses the control flow how each of these functions intend it to be used. Most events that are done with protractor will emit a promise that _can_ be acted on. 

The use of `done` allows us to be even more explicit as to when the test should notify the next test to start running. This can also be used in a `beforeAll` function:
```js
beforeAll(function(done) {
  browser.wait(EC.elementToBeClickable(filterDropDown), browser.params.defaultTimeout).then(function () {
    return filterDropDown.click()
  }).then(function () {
    return browser.wait(EC.visibilityOf(element(by.id('number-menu'))), browser.params.defaultTimeout);
  }).then(function () {
    return element(by.id('ui-id-2')).click();
  }).then(function () {
    return deriva.waitForElementInverse(element.all(by.id("loadIcon")).get(0));
  }).then(function () {
    done();
  }).catch(function (err) {
    console.log(err);
    done.fail();
  });
```

test/e2e/specs/treeview/00-static-treeview.spec.js
```diff
@@ -37,23 +37,132 @@ exports.tests = function (appName, baseUrl) {
...
+          it('node list expanded on + click', function() {
+            btn.click();
+            browser.wait(EC.visibilityOf(element(by.css('#nid_9112 > ul'))), browser.params.defaultTimeout);
+            element(by.css('#nid_9112')).all(by.tagName('li')).each(function (elm) {
+              expect((elm).isDisplayed()).toBeTruthy('Expanded List list is not visible.');
+            });
+          });
+
+          it('node list collapse on - click', function() {
+            btn.click();
```
jrchudy: This click event is sometimes triggering for me then failing with this message:
`element click intercepted: Element <i class="jstree-icon jstree-ocl" role="presentation"></i> is not clickable at point (33, 1238). Other element would receive the click: <i class="jstree-icon jstree-ocl" role="presentation"></i>`

I'm not sure what is wrong with this test case without picking it apart and really debugging it with code changes to the test.

test/e2e/specs/treeview/00-static-treeview.spec.js
```diff
@@ -37,23 +37,132 @@ exports.tests = function (appName, baseUrl) {
...
         });
+
+        describe('hierarchy', function() {
+          beforeAll(function() {
+            btn = element(by.css('#nid_9112 > i'));
+            deriva.waitForElement(btn);
+          });
+
+          it('node list expanded on + click', function() {
+            btn.click();
```
jrchudy: This click event is sometimes triggering for me then failing with this message:
`element click intercepted: Element <i class="jstree-icon jstree-ocl" role="presentation"></i> is not clickable at point (33, 1238). Other element would receive the click: <i class="jstree-icon jstree-ocl" role="presentation"></i>`

I'm not sure what is wrong with this test case without picking it apart and really debugging it with code changes to the test.

test/e2e/specs/treeview/00-dynamic-treeview.spec.js
```diff
@@ -262,20 +276,282 @@ exports.tests = function (appName, baseUrl) {
...
+                      }).catch(function (err) {
+                        console.log(err);
+                        done.fail();
+                      });
+                    });
+                 });
+            });
+            it('density Note should have correct tooltip', function() {
+                var densityNote = element(by.css('#nid_1431_anchor > span > span:nth-child(3) > img:nth-child(4)'));
+                densityNote.click();
```
jrchudy: Both places that you are clicking the density note icon, it _sometimes_ triggers before the modal dialog has finished fading out. I'm presented with this error:
`Failed: element click intercepted: Element  is not clickable at point (807, 401). Other element would receive the click: <div class="modal-backdrop fade"></div>`

You will have to play around with this a little bit, but I think you can use `elementToBeClickable` before clicking here. If that doesn't work, maybe checking that the modal element is no longer visible?

You could also try removing the modal animations so it just shows and dissappears rather than fading (this has helped with some of our other tests before).
```js
element(by.id("schematic-modal"))allowAnimations(false)
```

vipul-21:
@jrchudy I've updated the PR.

vipul-21:
PR merged after updating the code with reviewed changes.

PR #69, opened 2019-07-01
Jquery update to 3.4.1

vipul-21:

Issue #67, opened 2019-06-18
Make resource path based on chaiseBasePath property


jrchudy:
For both `boolean-search` and `heatmap`, the `chaiseConfig.js` that is used and all the rest of the chaise resources, are hardcoded to be `/chaise/...`. This means that the ones installed at the root of the server will be used. 

This is fine for the general deployments, but doesn't allow for testing of the integration between the chaise code and deriva-webapps code. The `chaiseConfig` should be fetched relative to the user's current directory (as well as the rest of the chaise assets).

This is going to be tricky, because the `chaiseBasePath` property is defined in the `chaiseConfig`, which is also in the `/chaise/` folder. So we need to think of how to make this path relative to get the `chaiseBasePath` from the `chaiseConfig`, then rely on the base path value for the rest of the `chaise` assets.

jrchudy:
To accomplish this, the deriva-webapps should follow a similar structure to this block of code in `navbar.app.js` in the chaise repo.

https://github.com/informatics-isi-edu/chaise/blob/master/lib/navbar/navbar.app.js#L66-L134

jrchudy:
During migration of both heatmap and boolean search to reactJS, this is solved by how we are using chaise `AppWrapper`

PR #70, opened 2019-07-29
Lineplot according to chaise guidelines

vipul-21:


vipul-21:
Note: The jquery used is 1.11.1 till the time we don't merge the [jquery upate PR](https://github.com/informatics-isi-edu/deriva-webapps/pull/69)

lineplot/lineplot.app.js
```diff
@@ -0,0 +1,231 @@
...
+                        };
+                        var tracesComplete = 0;
+                        lineplotConfig.traces.forEach(function (trace) {
+                            var uriWithFilters = baseUri + trace.path + "/subject_id=" + $rootScope.subject_id + "&recorded_time::geq::" + UriUtils.fixedEncodeURIComponent(timestamp);
+                            if ($rootScope.duration) {
+                                var end_x = moment(timestamp).add($rootScope.duration, 'h').format(dataFormats.datetime.submission);
+                                uriWithFilters += "&recorded_time::leq::" + UriUtils.fixedEncodeURIComponent(end_x);
+                            }
+                            var uri = uriWithFilters + "/" + trace.x_col + "," + trace.y_col + "@sort(recorded_time)?limit=" + $rootScope.limit;
+                            $http.get(uri).then(function(response) {
```
jrchudy: You should use the `$http` service that we have bundled in ermrestJS, [described here](https://github.com/informatics-isi-edu/deriva-webapps/wiki/Guideline-to-using-ermrestJS-error-handling-and-logging). This will ensure that the error handling in ermrestJS is used for the errors that are thrown.

For the service location, you should use the chaise config property, `ermrestLocation` if it's defined, then default to the default value of `<hostname>/ermrest`.

lineplot/lineplot.app.js
```diff
@@ -0,0 +1,231 @@
...
+                                plots.config = config;
+
+
+                                tracesComplete++;
+                                if (tracesComplete == lineplotConfig.traces.length) {
+                                    $rootScope.plots = plots;
+                                    Plotly.relayout();
+                                }
+                            }).catch(function (err) {
+                                if (err.data) {
```
jrchudy: You should use the error handling we have defined in chaise to handle this case. It looks like the behavior you are implementing is what we have in `chaise`.

lineplot/lineplot.app.js
```diff
@@ -0,0 +1,231 @@
...
+                $scope.showLineplot = function () {
+                	if (vm.model.user == null) {
+                		vm.model.user = $rootScope.user;
+                	}
+                    $scope.$apply(function () {
+                        vm.plotsLoaded = true;
+                    });
+                }
+
+                vm.logout = function () {
```
jrchudy: Why does this function need to be re-written? Was there an issue with using `Session.logout()`?

vipul-21:
PR updated with suggested changes.

lineplot/index.html
```diff
@@ -1,126 +1,86 @@
-<html ng-app="lineplotApp">
+<html id="lineplot">
 
 <head>
   <script src='/chaise/chaise-config.js'></script>
-  <script src='/chaise/scripts/vendor/jquery-1.11.1.min.js'></script>
+  <script src='/chaise/scripts/vendor/jquery-3.4.1.min.js'></script>
```
jrchudy: When I pull your changes and upload the code to my directory on prisms, I get an error that this file is not available. This current path assumes the file will be fetched from the root of the server instead of relative to the user's current location. 

For now, can you change the script tags in this html document to point to chaise relative to your current location? i.e. `../chaise/scripts/vendor/...` and `../chaise/common/...` for links further down the page. More info about why this is bad can be seen in issue #67 .

vipul-21: Updated it with relative paths.

jrchudy:
@vipul-21  since updating the script tags to use relative paths, did you then install the latest chaise code to your user directory and test the application?

vipul-21:
@jrchudy No I didn't update to the latest chaise code since the the use of relative paths.

PR #72, opened 2019-08-05
2D Plot

vipul-21:


plot/index.html
```diff
@@ -0,0 +1,69 @@
+<html id="plot">
+
+<head>
+  <script src='../../chaise/chaise-config.js'></script>
+  <script src='../../chaise/scripts/vendor/jquery-1.11.1.min.js'></script>
+  <script src='../../chaise/scripts/vendor/bootstrap-3.3.7.min.js'></script>
+  <script src='vendor/plotly-latest.min.js'></script>
+  <script src='vendor/moment.min.js'></script>
```
jrchudy: These 2 libraries are available in chaise as well, so you should include them based on the relative chaise path instead. Then remove them from the deriva-webapps/plot folder.

plot/README.md
```diff
@@ -0,0 +1,65 @@
...
+      1. uri : The url from which the data has to be fetched.
+      2. legend: The value of legend to be shown for this trace.
+      3. x_col: The column name for the x values
+      4. y_col: An array of column name for the y values
+6. plotlyButtonsToRemove: The button to be removed shown by plotly by defualt.
+
+#### Note
+If any of the above (Presentation or data) values is not mentioned, the app will throw an error. If you don't want to set a value, set it to `null`.
+
+### Sample lineplot-config.js (also included in the repo)
```
jrchudy: Thank you for adding an example in the readme 👍

plot/index.html
```diff
@@ -0,0 +1,69 @@
...
+  <script src='../../ermrestjs/ermrest.js'></script>
+  <script src="plot-config.js"></script>
+  <script src="plot.app.js"></script>
+  <link rel="stylesheet" type="text/css" href="styles/plot.css">
+  <meta name="viewport" content="width=device-width, initial-scale=1">
+</head>
+
+<body class="chaise-body">
+  <div class="configure-container" ng-app="chaise.configure-plotApp"></div>
+  <div class="mainContainer">
```
jrchudy: To be consistent with the rest of the chaise apps, this class should be `app-container`. 

We use `app-container` to denote the element that wraps the whole app (including the navbar while excluding the config app).

plot/index.html
```diff
@@ -0,0 +1,69 @@
...
+          <alerts alerts="ctrl.alerts"></alerts>
+      </div>
+      <div ng-if= "ctrl.model.title" class="row">
+      	<div id="bookmark-container" class="col-xs-12 meta-icons">
+      		<div id="title" class="pull-left">
+      			<h1 id="page-title"><span>{{ctrl.model.title}}</span></h1>
+      		</div>
+      	</div>
+      </div>
+      <div class="plot-container">
```
jrchudy: This tag should have an id and a class similar to our other apps as well. We should have `id="main-content"` and `class=<appname>-container` (which is what you do have already, `class="plot-container"`).

Following this tag, you should add the following code block. You want to add a fe more tags, move the alerts section down, and then include your div with `class="plot"`:
```
<div class="main-container>
  <div class="main-body">
    <div class="alerts-container">...</div>

    <div class="plot" ng-repeat="plot in plots">
      ...
    </div>
  </div>
</div>
```

vipul-21:
Updated with requested changes.

PR #78, opened 2019-09-25
Added log as a type and resolved safari issue

vipul-21:


plot/plot.app.js
```diff
@@ -164,16 +164,18 @@
...
                           title: plot.y_axis_label ? plot.y_axis_label : '',
                           automargin: true,
+                          type: config.y_axis_type ? config.y_axis_type : 'auto',
                       };
                       return layout;
                   }
                 }
 
-                function formatData(data) {
+                function formatData(data) { // HACK: for the demo data
```
jrchudy: Seeing as this function caused an issue earlier, let's change it to use try/catch.
```js
function formatData(data) {
  try {
    var formated_data = parseInt(data.split(' ')[0], 10);
    if (isNaN(formated_data)) {
      return data;
    }
  } catch (e) {
    return data;
  }
}
```

mikedarcy: This isn't really sufficient, because it will mangle string values which may have numbers in them and that are not really meant to have this special case apply.  A better approach would be to put in a config variable that only executes this behavior if `true`, otherwise this function is either bypassed entirely or is effectively a noop and just returns `data` unmodified.

plot/plot.app.js
```diff
@@ -174,16 +172,18 @@
...
                   }
                 }
-
-                function formatData(data) { // HACK: for the demo data
-                  var formated_data = parseInt(data.split(' ')[0], 10);
-                  if (isNaN(formated_data)) {
-                    return data
+                function formatData(data) {
+                  try {
+                    var formated_data = parseInt(data.split(' ')[0], 10);
```
mikedarcy: This fix still allows string data where spaces and integers are present to be arbitrarily reformatted, which should never be the default behavior. The execution of this function needs to be configurable at the plot level.

PR #80, opened 2019-10-03
2dplot

vipul-21:


plot/plot.app.js
```diff
@@ -191,26 +188,44 @@ function loadModule() {
                       if (isNaN(formated_data)) {
                         return data;
                       }
-                      return parseInt(formated_data.toLocaleString('en-US'));
+                      return  formated_data.toString().replace(/(\d)(?=(\d{3})+(?!\d))/g, '$1,')
```
jrchudy: Could you put a comment explaining what this `replace()` is doing

plot/plot.app.js
```diff
@@ -191,26 +188,44 @@ function loadModule() {
...
                     } catch (e) {
                       return data;
                     }
                   }
                   else {
-                    return data
+                    return data.replace(/(\d)(?=(\d{3})+(?!\d))/g, '$1,');
```
jrchudy: Could you put a comment explaining what this `replace()` is doing

plot/plot.app.js
```diff
@@ -191,26 +188,44 @@ function loadModule() {
...
-                            count+=1
-                        });
+                        var plotConfig = plotConfigs[config];
+                        var message = "";
+                        var error;
+
+                        if (plotConfig == undefined) {
+                          if (config) {
+                            message = "Invalid config parameter in the url";
+                            error = new Errors.CustomError("Invalid config", message);
```
jrchudy: Do we want this to error out? Or should we show a warning to the user and default to the "*" context (if defined)?

I think the case when the config doesn't match and "*" is not defined, should be when an error is thrown.

vipul-21: This was done for the case if the person specifies `config='something'` and that config is not present, then show the error( Instead of using the default config). 
If the config parameter is not defined in the query then use the default config.   
I'll check with Hongsuda about this.

vipul-21: @hongsudt Need your thoughts on this.

vipul-21: @jrchudy I had a chat with hongsuda, and she agrees with the way I implemented it.

Issue #86, opened 2020-04-02
plot: bugs 


hongsudt:
I found the following bugs trying to configure plots for rbk: 

- [x] title flashed and then disappear. I believed this is because the position of the title is incorrect. Also, please fix the flashing problem. 
- [x] pie chart: some of the higher values are missing e.g. Imaging:ISH or Imaging:IHC. Run the app with `config=gudmap-todate-pie` and compare it to `config=gudmap-todate-bar`; both display data from the same url. 
- [x] The app shouldn't be deployed with plot-config.js. We should mimic what chaise does by deploying the app with plot-config-example.js so people can use that as a template to start. The deployment specific config can them put their plot-config.js in the deployment for their own config.  
- [x] If plotly-config exists, the app seems to ignore `title`, `height`, `width` which are not defined as part of the `config` attribute. The only thing that seems to work is the plot type. We did discuss if plotly-config exist, its config wins.  
- [x] It seems like the attribute for processing the labels (x-column) are different between the pie chart and bar chart (`format_data` v.s. `format_data_x`). Shouldn't it be using the same attribute name? 
   - for pie chart, there is no "x" or "y" notion in the data, so the configuration doesn't make that distinction either
- [ ] We should change the format_data feature to present the proper label for sequencing data e.g. for 100 (20/80), 100 should be used for plot but the label for the y-value should still be 100 (20/80). 
- [x] Make the x or y labels linkable to chaise pages
- [ ] Support Pre and Post plot that will take markdown, so we can add a link to switch from plot to table?
- [ ] #120

PR #87, opened 2020-04-22
Title flashing issue on load

vipul-21:
This PR resolves the issue 1 mentioned in the #86


jrchudy:
Looks like this fixed the issue with the title js text flashing before disappearing again.

PR #88, opened 2020-05-04
plot config sample as default

vipul-21:
The app shouldn't be deployed with plot-config.js. We should mimic what chaise does by deploying the app with plot-config-example.js so people can use that as a template to start. The deployment specific config can then put their plot-config.js in the deployment for their own config.


vipul-21:
Yes, I intend for only sample to be included in the repository and the actual config to uploaded at build time.
The bugs reported in this #86, it addresses the 3rd point.

jrchudy:
Sounds good. So lets add `plot/plot-config.js` to `.gitignore` so the file isn't included in the repository.

PR #89, opened 2020-05-04
plot-app pie chart not showing the comma separated values

vipul-21:
Pie chart: Was not able to render comma separated values. The issue is mentioned in #86 as second bug.


jrchudy:
How can I go about testing this fix from the changes in this PR?

vipul-21:
The pie chart in this [link](https://www.gudmap.org/deriva-webapps/plot/?config=gudmap-todate-pie) on prod does not render the values having commas in them.
To use the same config as this i.e `config=gudmap-todate-pie`, you can use the config from [here](https://www.gudmap.org/deriva-webapps/plot/plot-config.js) and see the values greater than 1000 are being displayed in the pie chart.

jrchudy:
Ok cool. I copied that config to my local directory and see the changes.

PR #91, opened 2020-07-30
Enable treeview background in print mode and persist scrolling

DamCousik:
This PR includes the following fixes

- [x] Fixed the print for iframes with Treeview document to show the tree hierarchy during print.
- [x] Persisted scrolling during print
- [x] Parsing the URL to check if there is an embedded YouTube video. If yes, then add a class to apply the above discussed custom styles.
- [x] The same has been processed for the <video> markdown pattern as well.

- Applying the thumbnail to the YouTube video on print was complicated as the styles are applied via YouTube and we have limited access to it. Hence, the alternative was to display a note informing the user that the video is hidden during print and possibly provide the URL of the hidden video.

### DIFF
--- a/treeview/index.html
+++ b/treeview/index.html
@@ -6,7 +6,7 @@
     <!-- issue with using out own jquery-ui css is that it will be missing icons/images that have a relative path on jquery website -->
     <!-- <link rel="stylesheet" href="resources/styles/jquery-ui.css"> -->
     <link rel="stylesheet" href="https://code.jquery.com/ui/1.12.1/themes/base/jquery-ui.css">
-    <script src="../../chaise/scripts/vendor/jquery-3.3.1.min.js"></script>
+    <script src="../../chaise/scripts/vendor/jquery-3.4.1.min.js"></script>
     <script src="util/jquery-ui.js"></script>
     <script src="util/jstree.js"></script>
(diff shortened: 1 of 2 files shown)


DamCousik:
@RFSH / @jrchudy : I also tried to check if I could bypass the browser settings to always display the background images but I was unable to. I tried the following options as suggested in this [post](https://stackoverflow.com/questions/6670151/how-can-i-force-browsers-to-print-background-images-in-css) but had no luck!

1) Added the changes to `<body>`

```
body {
    content:url(32px.png) !important;
    -webkit-print-color-adjust: exact !important;   /* Chrome, Safari */
    color-adjust: exact !important;                 /*Firefox*/
  }
```

2) Added it to the universal selector *

```
/*To force the browser to print background images and colors*/
  * {
    -webkit-print-color-adjust: exact !important;   /* Chrome, Safari */
    color-adjust: exact !important;                 /*Firefox*/
  }
```

Since nothing worked, I haven't committed the changes. I also ready a couple of articles and in most of them, the discussion say we have very little control over the browser's printing capabilities. However, I found this [post](https://rewordify.com/BackgroundColors.htm) which explains how we can turn on the settings in all major browsers to have the background images printed.

PR #94, opened 2020-08-14
Make install no config

jrchudy:
This PR only addresses the first bullet in the connected issue:
> Plot config should not be added to build folder from the repo. Remove from make install command

1. Changed the `<app>-config.js` files that existed in the folders to `<app>-config-sample.js`
2. Modified `make install` command to exclude adding `<app>-config.js` and `<app>-config-sample.js` to the server folder when run.
3. changes `<head>` in `index.html` for each app to do the following:
    ```HTML
    <script src="<app>-config-sample.js"></script>
    <script src="<app>-config.js"></script>
    ```
   - I was uncertain about this but didn't want to break functionality in treeview that is relying on the sample config right now. I wanted a way to provide a configuration to be placed in the folder that isn't in github to override the sample that is provided so sample will be used by default. If another config is placed in the app's folder, it will be used instead of the default.


boolean-search/index.html
```diff
@@ -28,6 +28,7 @@
     <script src='../../chaise/common/config.js'></script>
     <script src='../../ermrestjs/ermrest.js'></script>
     <link rel="stylesheet" type="text/css" href="styles/booleansearch.css">
+    <script src="booleansearch-config-sample.js"></script>
```
RFSH, line 31: I know that you said this didn't break anything, but this doesn't look correct. We shouldn't use the sample file and sample file is just an example that can be used to create the actual config files.

.gitignore
```diff
@@ -1,4 +1,8 @@
 node_modules
 .DS_Store*
+plot/booleansearch-config.js
```
RFSH, line 3: These values are incorrect (shouldn't start with plot)

plot/plot-config.js
```diff
@@ -1,311 +1,370 @@
 var plotConfigs = {
```
RFSH, line 1: you should remove this `plot-config.js` file as well (like the other config files)

PR #131, opened 2023-02-16
Migration of the matrix into react

cheekw:
New features:
- Virtualized grid for optimized performance
- Search bar with virtualized options list
- New selectable colormaps with colorblind friendly options
- Dynamically sized grid
- Grid scrolling buttons
- Row / Column Headers always show
- Legend always shows
- Config dictates sizing
- Ellipses to truncate long headers and labels


src/components/virtualized-select.tsx
```diff
@@ -0,0 +1,95 @@
+import { memo, useState } from 'react';
+import Async, { AsyncProps } from 'react-select/async';
+import { MenuListProps, StylesConfig } from 'react-select';
+import { FixedSizeList } from 'react-window';
+import { SelectComponents } from 'react-select/dist/declarations/src/components';
+
+type VirtualizedSelectProps = AsyncProps<any, any, any> & {
```
RFSH, line 7: Same comment that I left for search-bar.tsx about separating `AsyncProps` and your component props.

RFSH (changes requested):
Apart from the comments that I left for different parts of the code, you need to add comments to each components and its props.

I'm not going to mention each individual component or function, so please go over all the functions/components/props that you have and add comments. Also if a function is complicated and is doing multiple things, please add inline comments summarizing what it's doing.

hooks/matrix.ts
```diff
@@ -0,0 +1,269 @@
...
+        }
+      });
+      return result;
+    },
+    [colorThemeOption.value]
+  );
+
+  // Side Effect for Updating Data
+  useEffect(() => {
+    const fetchMatrixData = async (config: any) => {
```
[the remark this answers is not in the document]
cheekw: This is to fetch the data. I decided to use async/await notation so that once the data is fetched asynchronously, I can set the state synchronously.

src/components/virtualized-select.tsx
cheekw: I think this one might be fine as the only component I am returning in this component is the Async itself. It is just modifying it. I have changed it in the search-bar.js as suggested though.

hooks/matrix.ts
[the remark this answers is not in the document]
RFSH: Right. What I meant is that you don't need to define a function and just have the content of function below. This function is called only once and it's easier to read it that way.

So line 236:
```ts
try {
  const config = getConfigObject(matrixConfigs);

  // the content of this function directly should be here:
  const xPromise = ConfigService.http.get(config.xURL);
  ....

 

} catch () {
 //....
} 
```

src/components/matrix/column-headers.tsx
```diff
@@ -0,0 +1,98 @@
+import { memo, forwardRef, ForwardedRef, CSSProperties } from 'react';
+import { VariableSizeList as List, ListOnScrollProps } from 'react-window';
+
+type ColumnHeadersProps = {
+  cellHeight?: number; // height of grid cell
```
RFSH, line 5: Use jsdoc style like I mentioned.

```ts
type ColumnHeadersProps = {
  /**
   * height of grid cell
   */ 
  cellHeight?: number;
  ....
}

```

The goal is using a standard way of defining comments, so VSCode or any other tools can pick up these comments and show them.

src/components/matrix/column-headers.tsx
```diff
@@ -0,0 +1,98 @@
...
+  type HeaderComponentProps = {
+    index: number;
+    data: any;
+    style: CSSProperties;
+  };
+
+  /**
+   * Header component for each column
+   */
+  const HeaderComponent = ({ index, data, style }: HeaderComponentProps): JSX.Element => {
```
RFSH, line 47: I'm not familiar with how this works. Why is this component defined inside the other component? If it doesn't have to be, move it to be outside so it's easier to read.

src/components/matrix/grid-cell.tsx
```diff
@@ -0,0 +1,89 @@
+import { memo, CSSProperties } from 'react';
+import { areEqual } from 'react-window';
+
+export type GridCellProps = {
+  columnIndex: number;
```
RFSH, line 5: As I said I'm not going to comment on each individual component and I expect you to comment what each prop is. Even if it looks obvious, please add it for completeness. We might also eventually add more less obvious props, and that would allow us to consistently comment everything.

hooks/matrix.ts
```diff
@@ -0,0 +1,295 @@
...
+
+/**
+ * Hook function to use matrix data given a config object
+ *
+ * @param matrixConfigs
+ * @returns all data to be used by matrix visualization
+ */
+export const useMatrixData = (matrixConfigs: any) => {
+  const { dispatchError, errors } = useError();
+  const [styles, setStyles] = useState<any>(null); // styles from the config object
```
RFSH, line 178: Move the comments about the `useState` line to make it easier to read for lengthier lines.

hooks/matrix.ts
```diff
@@ -0,0 +1,295 @@
...
+  { value: 'viridis', label: 'Viridis' },
+];
+
+/**
+ * Hook function to use matrix data given a config object
+ *
+ * @param matrixConfigs
+ * @returns all data to be used by matrix visualization
+ */
+export const useMatrixData = (matrixConfigs: any) => {
```
RFSH, line 176: Create a type that describes the returned object. This would also help you realize that some of the stuff that you're returning are not used and you should remove them.

For example you can see how when you use `useError`, it will tell you the type of each prop and their purpose. That's the goal here.

hooks/matrix.ts
```diff
@@ -0,0 +1,295 @@
...
+  };
+};
+
+/**
+ * Generate a link for combination of x and y filter on the main table
+ * @param  {object} x the x object, must have .title and .id
+ * @param  {object} y the y object, must have .title and .id
+ * @return {string} returns the url string to be used
+ */
+const generateLink = (config: any, x?: any, y?: any, z?: any) => {
```
RFSH, line 272: Why is this function down here, and the other "util" functions above the hook? Either move this to be above the `useMatrix`, or move the other function to be after this function.

hooks/matrix.ts
RFSH: With this you would also have to type `matrixData` which should be straightforward given that you're creating the object from scratch.

config/matrix-config-sample.js
```diff
@@ -1,21 +1,31 @@
...
+    zFacetColumn: 'id',
     catalogId: '1',
     schemaName: 'isa',
     tableName: 'dataset',
     prularDataset: 'datasets',
-    matrixContainer: '.mousematrix-container'
-  }
+    matrixContainer: '.mousematrix-container',
+    styles: {
+      title: 'Mouse Data Summary',
```
RFSH, line 19: - Are you using `matrixContainer` and `prularDataset`? If not, please remove them.

- I don't think title and subtitle in "styles" makes sense. Let's just have them in the first level and use `title_markdown` and `subtitle_markdown`.

config/matrix-config-sample.js
```diff
@@ -1,21 +1,31 @@
...
+      subtitle: 'See at a glance all experiment types for our mouse data by anatomy and age stage',
+      maxCols: 30,
```
RFSH, line 21: I would have to create a documentation explaining each property. Please add comments to each line in here so I would know what each property is. You should also explain what happens if the property is not defined. The whole "styles" object should be optional by the way.

For example:

```js
/**
 * maximum number of rows to display by default
 * if this is not defined, we will fit as many as the viewport size allows
 */
maxRows: 100,
/**
 * Height of each cell in pixel. 
 * Default value: 25
 */
cellHeight: 25,

```

hooks/window-size.ts
```diff
@@ -0,0 +1,30 @@
+import { useState, useEffect } from 'react';
+
+export type WindowSize = {
+  width: number | undefined;
```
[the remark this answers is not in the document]
RFSH: You haven't commented it. Always use jsdoc style so the IDEs can understand it.

src/pages/matrix.tsx
```diff
@@ -1,62 +1,246 @@
...
+    height - colHeaderHeight - heightBufferSpace, // can't exceed browser height
+    colHeaderHeight + cellHeight * maxRows
+  );
+  const gridWidth = Math.min(
+    cellWidth * numColumns, // can't exceed total grid
+    width - rowHeaderWidth - widthBufferSpace, // can't exceed browser width
+    rowHeaderWidth + cellWidth * maxCols
+  );
+
+  const gridTitle = styles.title;
```
RFSH, line 161: As I mentioned we want to use `title_markdown` and `subtitle_markdown`. You should make sure you're handling where these two are missing/invalid.

And to parse markdown you would do the following:

```ts
const title = ConfigService.ERMrest.renderMarkdown(title_markdown);

 // add title only if it's defined
if (title && title.length > 0) {

}
```

hooks/matrix.ts
```diff
@@ -0,0 +1,295 @@
...
+      const xyzPromise = ConfigService.http.get(config.xysURL);
+
+      // Batch the requests in Promise.all so they can run in parallel:
+      const data = await Promise.all([xPromise, yPromise, zPromise, xyzPromise]);
+      const parsedData = parseMatrixData(config, data);
+
+      // Set state after the request completes
+      setData(data);
+      setMatrixData(parsedData);
+      setStyles(config.styles);
```
RFSH, line 230: You should handle the cases when `styles` is completely missing. Or when some properties are not defined. Your code is assuming that all the props are always defined.

src/components/matrix/virtualized-grid.tsx
```diff
@@ -0,0 +1,408 @@
...
+          d='M12.354 1.646a.5.5 0 0 1 0 .708L6.707 8l5.647 5.646a.5.5 0 0 1-.708.708l-6-6a.5.5 0 0 1 0-.708l6-6a.5.5 0 0 1 .708 0z'
+        />
+      </svg>
+    </button>
+  );
+};
+
+const GridUpButton = ({ onClick, rowHeaderWidth }: GridMoveButton): JSX.Element => {
+  return (
+    <div
```
RFSH, line 282: I think you mentioned that you've borrowed these from bootstrap icons. So please add a link to the source

src/pages/matrix.tsx
```diff
@@ -1,62 +1,246 @@
...
-    try {
-      const config = getConfigObject(windowRef.matrixConfigs);
-      setMatrixProps({ config });
-
-    } catch (error: any) {
-      dispatchError({ error })
-    }
-
-  }, []);
+  const { width = 0, height = 0 } = useWindowSize();
```
RFSH, line 32: Let's define these default values in the hook itself.

RFSH (changes requested):
Overall your code looks great. It's readable, and I couldn't find any major issues. The main thing that you should work on is typing and commenting. When I mentioned commenting on the types, you should follow the jsdoc style. That way VSCode will understand the comment and will show these comments when you're using a hook or component.

You should properly type the returned value of `useMatrix`. You're creating the data from scratch, so adding types should be straightforward. It will also make you realize that some properties are unused and should be removed.

Apart from commenting and types,

- `styles` and their properties should be optional. So make sure you're properly handling it in `useMatrix` hook.
- Use `title_markdown` and `subtitle_markdown` as I mentioned in my comments.

hooks/matrix.ts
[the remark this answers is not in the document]
RFSH: After discussing this, I realized that this is the typical style of doing async calls in `useEffect`.

src/pages/matrix.tsx
```diff
@@ -1,62 +1,246 @@
...
+  return (
+    <div className='matrix-page'>
+      {toastMessage ? (
+        <div className='alert alert-danger' role='alert' style={toastStyles}>
+          {toastMessage}
+        </div>
+      ) : null}
+      <div className='content-container'>
+        <div className='title-container'>
+          <h1>{gridTitle}</h1>
```
RFSH, line 197: Assuming you've changed `gridTitle` to be the output of `renderMarkdown`, then this line should be

```jsx
<h1><DisplayValue addClass value={{value: title, isHTML: true}} /></h1>
```

hooks/window-size.ts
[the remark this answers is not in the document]
cheekw: left as is as discussed on slack but added comments.

src/pages/matrix.tsx
cheekw: I think this one is ok behavior. if it's undefined it means something else such as no window size exists e.g. executed server side: https://usehooks.com/useWindowSize/

hooks/matrix.ts
cheekw: handled by changing to config

src/pages/matrix.tsx
cheekw: modified

hooks/matrix.ts
cheekw: typed!

src/components/matrix/grid-cell.tsx
cheekw: added for most exported types for now

src/components/matrix/column-headers.tsx
cheekw: moved outside

hooks/matrix.ts
cheekw: I think typically state comments are not that fleshed out. I typically don't see any more comments than this for useState. 

More complicated state can be built into a reducer and documented on that instead.
I have moved it to the preferred style though.

src/components/virtualized-select.tsx
```diff
@@ -0,0 +1,132 @@
...
+   */
+  value: string;
+  /**
+   * displayed label of option
+   */
+  label: string;
+};
+
+type VirtualizedSelectProps = AsyncProps<any, any, any> & {
+  itemHeight: number; // the height of each menu item
```
RFSH, line 19: Use jsdoc style of comments

RFSH (approved):
Thank you for all the changes. The PR is ready to be merged.

PR #134, opened 2023-03-09
React plot

cheekw:
- cleaned up matrix a bit
- added config typing
- react migration with functional bar, pie, histogram, scatter


src/components/matrix/grid-button.tsx
```diff
@@ -0,0 +1,183 @@
...
+
+export type GridMoveButton = {
+  onClick: MouseEventHandler;
+  rowHeaderWidth: number;
+};
+
+/**
+ * SVG Icon for button from https://icons.getbootstrap.com/
+ */
+export const GridLeftButton = ({ onClick, rowHeaderWidth }: GridMoveButton): JSX.Element => {
```
cheekw, line 11: @RFSH moved these buttons to a separate file to make the grid component cleaner

src/hooks/matrix.ts
```diff
@@ -160,7 +162,7 @@ type MatrixData = {
   /**
    * matrix configurations
    */
-  config: any;
+  config: MatrixDefaultConfig | null;
```
cheekw, line 165: added typing @RFSH

src/models/matrix-config.ts
```diff
@@ -0,0 +1,44 @@
+export type MatrixDefaultConfig = {
```
cheekw, line 1: Matrix config typing @RFSH

package.json
```diff
@@ -26,6 +29,7 @@
   },
   "dependencies": {
     "@isrd-isi-edu/chaise": "0.0.8",
+    "plotly.js-cartesian-dist-min": "^2.18.2",
```
cheekw, line 32: Added cartesian to support plotly plots 

@jrchudy

src/utils/string.ts
```diff
@@ -0,0 +1,144 @@
...
+};
+
+/**
+ * Creates the link bemplateased on the given markdown pattern and templateParam
+ *
+ * @param markdownPattern pattern used to create link 
+ * @param templateParam params used to replace values in the pattern to create the link
+ * @returns
+ */
+export const createLink = (markdownPattern: string, templateParam?: any): string => {
```
cheekw, line 112: Created some helper functions that were often getting used in plot.app.js

@jrchudy

types/plotly.js-cartesian-dist-min.d.ts
```diff
@@ -0,0 +1,4 @@
+declare module 'plotly.js-cartesian-dist-min' {
```
cheekw, line 1: for typing

src/models/matrix-config.ts
```diff
@@ -0,0 +1,44 @@
...
+  xURL: string;
+  yURL: string;
+  zURL: string;
+  xysURL: string;
+  xFacetColumn: string;
+  yFacetColumn: string;
+  zFacetColumn: string;
+  xSource: AxisSource;
```
RFSH, line 9: Let's use `any` for the `xSource`, `ySource`, and `zSource`. Its value could be a string or array of elements. The typing should be something that we're eventually going to add to ermrestjs.

RFSH (approved):
The matrix changes look good to me.

jrchudy (approved):
This looks really good. I tested each example in the github issue I created and can see each working as I expect. The only inconsistency is the location of the legend by default for pie charts, but as we discussed on Wednesday that seems to be something in plotly that has changed.

Only question I had was regarding using the `plotly.js-cartesian-dist-min` library. Was this because not all types we want to support  were supported in the basic version of the library? I imagine violin might be one type that wasn't supported in the basic library.

PR #136, opened 2023-03-10
React matrix search width bug

### LINKED ISSUE #133: Matrix search bar is cutting content in Safari
On Safari 16.3. when you first load the page, and type a search (e.g., "mandible") at some point while typing the word, the text in the search box starts to scroll to the left as if there wasn't enough room in the search box control for the text. 

This is only happening on the first search attempt and after that the search box properly adjusts the content. It's also not reproducible in Firefox or Chrome. 

Screen Shot 2023-02-24 at 11 29 22 AM

cheekw:
- removed excessive memoization
this fixes #133

### DIFF
--- a/src/components/search-bar.tsx
+++ b/src/components/search-bar.tsx
@@ -103,4 +103,4 @@ const SearchBarButton = ({ onClick }: SearchBarButtonProps): JSX.Element => (
 
 const MemoizedSearchBarButton = memo(SearchBarButton);
 
-export default memo(SearchBar);
+export default SearchBar;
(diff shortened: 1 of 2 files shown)


RFSH:
@cheekw hmm, do you know why `memo` might have caused this?

cheekw:
> @cheekw hmm, do you know why `memo` might have caused this?

memo() basically skips re-rendering unless props change.
But it doesn't do a deep equality check. Since we were wrapping SearchBar with memo() and were passing values through an object `selectProps`, it assumes that props haven't changed, causing it to not re-render.

RFSH (approved):

Issue #138, opened 2023-04-10
Clickable legend and plot in React plot app


jrchudy:
With the primary functionality for each plot type of plot app migrated, we need to make sure the previous functionality of the legend click event and graphic click event is being supported. 

For changing the legend click event, we want to support the `legend_markdown_pattern` and `disable_default_legend_click` properties. More information can be found in the [README](https://github.com/informatics-isi-edu/deriva-webapps/tree/master/plot).
 - `legend_markdown_pattern` - implemented as a **string array** that are used for changing the displayed value and also allowing data modelers to create the text as a link. `plots[x].traces[x].legend_markdown_pattern`
 - `disable_default_legend_click` - implemented as a boolean property to disable the default legend click event. `plots[x].config.disable_default_legend_click`
   - for example, the legend for pie charts allow the user to click a legend property. Single click will hide the pie slice associated with the clicked legend text. Double click will show only the pie slice associated with the clicked legend text
   - with this property set to `true`, the default single and double click events should be turned off
 - These 2 properties should work for the following types to maintain previous functionality:
   - [x] pie
   - [x] bar
   - [x] violin
 - make sure `disable_default_legend_click` also works for:
   - [x] dot
   - [x] histogram

For changing the graph click event, we want to support `graphic_link_pattern`. More information can be found in the [README](https://github.com/informatics-isi-edu/deriva-webapps/tree/master/plot).
 - `graphic_link_pattern` - implemented as a **string array** or a just a **string** to allow for graph click events to navigate to other pages. `plots[x].traces[x].graphic_link_pattern`
 - this property should work for the same types listed above:
   - [x] pie
   - [x] bar
   - [x] violin

NOTE:
For violin plot, the location of the configuration values differs from the other plot types.
 - `plots[x].config.xaxis.group_keys[x].legend_markdown_pattern`
 - `plots[x].config.xaxis.group_keys[x].graphic_link_pattern`

The implementation for allowing customized click events for the legend and graph was initially discussed in issue #106 . More details can be found there and in the [plot functionality document](https://github.com/informatics-isi-edu/deriva-webapps/blob/master/docs/user-docs/plot-functionality.md#legend).

Future implementation of these properties (`graphic_link_pattern` and `legend_markdown_pattern`) should be expanded to `dot` and `histogram`.

Issue #135, opened 2023-03-09
Migrate Violin plot type to react


jrchudy:
Violin plot is the next step for migrating the plot app. The complexity comes with implementing the selectors that show at the top of the page. These include the gene selector, study selector, group by dropdown, and scale dropdown. This is explained in some detail [here](https://github.com/informatics-isi-edu/deriva-webapps/blob/master/docs/user-docs/plot-functionality.md#violin-plot).

There are 3 modes the app can operate in, standalone app, embedded in study page, and embedded in gene page. The selectors behave slightly differently in each case.

#### standalone app
In this case, there is either nothing in the url ([example](https://dev.atlas-d2k.org/deriva-webapps/plot/?config=study-violin)) or both Study and NCBI_GeneID url parameters are present ([example](https://dev.atlas-d2k.org/deriva-webapps/plot/?config=study-violin&Study=14-3PXT&NCBI_GeneID=1)). 
 - All selectors should be available
 - with no url parameters:
   - fetch gene selector with no constraints
     - default to first gene in set returned
   - fetch study selector info based on gene
     - All Studies
   - default group to Anatomical Source (plot-config)
 - with both parameters
   - fetch gene selector info based on study
     - modify gene URL to append NCBI_GeneID
     - use gene from url as default
   - fetch study selector info based on gene
     - modify study URL to append RID
     - fetch tuple for study from url for rowname info to display
 
#### embedded in study page
In this case there is only Study defined in the url ([example](https://dev.atlas-d2k.org/deriva-webapps/plot/?config=study-violin&Study=14-3PXT)).
 - fetch gene selector info based on study
   - default to first gene in set returned
 - remove study selector (don't fetch study set)
   - fetch study info in case it's used in title or other parts of graph
   
#### embedded in gene page
In this case there is only Study defined in the url ([example](https://dev.atlas-d2k.org/deriva-webapps/plot/?config=study-violin&NCBI_GeneID=1)).
- disable gene selector (don't fetch gene set)
  - fetch rowname for gene that was selected and display in input
  - fetch study selector info based on gene
    - All Studies
   - default group to Anatomical Source (plot-config)
   
#### recordset modal
[...]

Issue #130, opened 2023-02-07
Migrate Plot App to React


jrchudy:
In this repository we have a pre-existing app we called `Plot App` in the `plot` folder. This app handles multiple basic `plot_type`s and their configuration that can be read about more in the [README](https://github.com/informatics-isi-edu/deriva-webapps/blob/master/plot/README.md). To begin, let's start with some of the simpler plot types. 

### Where to start
The `plot_type`s we should start with are `bar`, `pie`, `scatterplot`, and `histogram`. In the plot folder, we have a [sample configuration file](https://github.com/informatics-isi-edu/deriva-webapps/blob/master/plot/plot-config-sample.js) with examples for each of the above plot types. This configuration file should be copied and renamed `plot-config.js` then deployed alongside the plot code in your user directory on the server (similar to what was done for matrix).

Once deployed, each plot can be accessed similar to matrix by changing the `config=` query parameter to point to a different configuration from `plot-config.js`:
```
https://dev.atlas-d2k.org/deriva-webapps/plot/?config=gudmap-todate-bar
```

Examples:
- [bar](https://dev.atlas-d2k.org/~jchudy/deriva-webapps/plot/?config=gudmap-todate-bar)
- [pie](https://dev.atlas-d2k.org/~jchudy/deriva-webapps/plot/?config=gudmap-todate-pie)
- [scatterplot](https://dev.atlas-d2k.org/~jchudy/deriva-webapps/plot/?config=specimen-scatterplot)
- [histogram](https://dev.atlas-d2k.org/~jchudy/deriva-webapps/plot/?config=specimen-histogram-horizontal)

All of the existing code for the plot app is written in `angularJS` and uses the `plotly` library for rendering plot content. This can be found in `plot/plot.app.js`. You'll notice there is branching logic in there for fetching data. This is done the same way for most basic `plot_type`s and slightly differently for `violin-plot` since that relies on fetching data from multiple different APIs that we will handle later.

### First steps
Begin with creating a similar page file in `src/pages` to what was done for matrix (`src/pages/plot.tsx`). Then follow what was done with components as well `src/components/plot/`.

Basic functionality of plots can be read about in the [user-docs](https://github.com/informatics-isi-edu/deriva-webapps/blob/master/docs/user-docs/plot-functionality.md).

PR #140, opened 2023-04-19
React plot migration

cheekw:
#130 #135 #138 
- migrate components of plotly to react
- added dynamic select components implementation
- added use of react-based chaise components
- added support for multiple charts
- added support for multiple types of plots in a chart
- added support for fetch select changes and non-fetch select changes
- added parallelism in requests for initial fetch for filling selector data

may need to revisit templateParams
may need to look at the use of noData hack


jrchudy (changes requested):
From reviewing the UI functionality, I found some issues that need to be addressed before we can merge this pull request.

For violin plot:
 - [x] The gene selector looks disabled when it is NOT disabled
 - [x] When `Study` is in the url, the study selector should not be shown
    - check the issue #135  for more info. The study information for the specified study should still be fetched in case it's used for templating
  - [ ] When `NCBI_GeneID` is in the url, the gene selector should be disabled with the rowname of the gene from the url displayed
    - check the issue #135  for more info
    - input should be simple with no button and greyed out
  - [ ] When `Study` or `NCBI_GeneID` or both are present in the url, have a link in the top right corner to go to violin plot with no `Study` or `NCBI_GeneID` url parameters
  - [ ] `xaxis.group_keys[x].title_display_pattern` displays for the x axis title properly but this is also included in the group by dropdown display. In that case it's not being templated and showing as a raw string like `[Experiment](/chaise/recordset/#2/RNASeq:Experiment)`
    - we should not be showing this in the dropdown
    - ```
       group_keys: [
         {
           column_name: "Experiment",
           title_display_pattern: "[Experiment](/chaise/recordset/#2/RNASeq:Experiment)",
           ...                     
         }, ...
         ```
    - think about how this can be addressed and we will have to talk about what solution is best
  - [x] `graphic_link_pattern` doesn't work for violin plot
  
  
For plot app in general:
 - [ ] when the page is resized, the app should be responsive
   - when I open the browser console in the old version of the app, the plot shrinks to make sure the content can be shown in the visible part of the viewport
 - [x] `disable_default_legend_click` doesn't seem to be implemented
   - This attribute should be used to allow for turning off the default click functionality of the legend.
   - this default functionality should not be turned off just because there is a link defined, we want the config property to control this

cheekw:
Following fixes/changes were added:

Selectors:
- Study Selector gone when in URL
- NCBI_GeneID Selector gone when in URL
- fixed gene selector looking disabled

Violin Graphic Click:
- fixed violin graphic links, clickable on the violin plot.

Layout responsive behavior:
- absolute min width x height = 800 x 600, 
- responsive behavior VP smaller than given layout but bigger than min

Legend Click:
- fixed disable_default_legend_click

RFSH:
@cheekw I updated Chaise to the issues that I observed in plot app, so I pushed a change to this branch to use it. It should solve the race condition that we were seeing on recordset modal (the terminal error that sometimes happened). 

While testing I found two issues that I think you should address:
- In dev mode, React is complaining about usage of `button` inside another `button`. 
- In prod mode, I'm seeing duplicated catalog and schema requests to the server (In dev mode you will see 4).  This can happen if you're sending some requests in `useEffect` and aren't guarding against doing it only once.  See if you can what's triggering these requests and make sure all of your `useEffect`s are only called once. The catalog and schema requests are done as part of "introspection". So they could be triggered by `ERMrest.resolve` which is most probably what you're doing. `AppWrapper` is also calling the catalog request on load.

cheekw:
> @cheekw I updated Chaise to solved the issues that I observed in plot app, and pushed a change to this branch to use the new version of Chaise. It should solve the race condition that we were seeing on recordset modal (the terminal error that sometimes happened).
> 
> While testing I found two issues that I think you should address:
> 
> * In dev mode, React is complaining about usage of `button` inside another `button`.
> * In prod mode, I'm seeing duplicated catalog and schema requests to the server (In dev mode you will see 4).  This can happen if you're sending some requests in `useEffect` and aren't guarding against doing it only once.  See if you can find what's triggering these requests and make sure all of your `useEffect`s are only called once. The catalog and schema requests are done as part of "introspection". So they could be triggered by `ERMrest.resolve` which is most probably what you're doing. `AppWrapper` is also calling the catalog request on load.

- Button inside a button occurs because I have wrapped the disabled dropdown of the gene selector to disable the dropdown behavior and "mimic" a button. While this is usually bad practice, I think it could be fine here since the buttons inside are not active. It also saves time of building a "fake dropdown component"
- Multiple requests:
If you look closely at the request uri are different for the first 3. The 4th one was a bug that I just fixed and pushed.
Request 1: For Gene Selector (using resolve())
Request 2: For Study Selector (using resolve())
Request 3: The actual plot data (using http.get())
Request 4: (bug that has been fixed in the recent commit, it was doing an extra fetch on clicking a selector)

There maybe a way to bundle request 1,2,3 together. I haven't looked into if this was possible though. Requests 1 and 2 should be occurring in parallel to save time. 3 should be waiting on 1 and 2.

jrchudy:
I updated my above comment with checkboxes to track which tasks are complete. After reviewing the UI again, looks like most of the concerns have been fixed. The gene selector was changed but not fixed properly and 2 other things were not addressed (listed in above comment).

The issue (#135) describes how the app is supposed to behave when 1 of `Study` or `NCBIGene_ID` are present in the URL, when both are present, and neither are present.

No url parameters:
 - https://staging.gudmap.org/deriva-webapps/plot/?config=study-violin
 - show both selectors active

Both `Study` and `NCBIGene_ID`:
 - https://staging.gudmap.org/deriva-webapps/plot/?config=study-violin&Study=14-3PXT&NCBI_GeneID=1
 - show both selectors active
 - show `[Visualization with Gene & Study selectors](https://staging.gudmap.org/deriva-webapps/plot/?config=study-violin)` link in top right

Only `Study`:
 - https://staging.gudmap.org/deriva-webapps/plot/?config=study-violin&Study=14-3PXT
 - hide study selector
 - show `[Visualization with Gene & Study selectors](https://staging.gudmap.org/deriva-webapps/plot/?config=study-violin)` link in top right
 
Only `NCBIGene_ID`:
 - https://staging.gudmap.org/deriva-webapps/plot/?config=study-violin&&NCBI_GeneID=1
 - disable gene selector
 - show `[Visualization with Gene & Study selectors](https://staging.gudmap.org/deriva-webapps/plot/?config=study-violin)` link in top right

jrchudy (commented):
Code changes look good. Please address the other minor issues I mentioned and we can finalize the code review to merge these changes.

Regarding the button in button that you mentioned in the other comment, we have a similar implementation in chaise, that can be found [here](https://github.com/informatics-isi-edu/chaise/blob/master/src/components/input-switch/foreignkey-field.tsx#L206-L235). We have a div marked as a `chaise-input-group` which is a wrapper we made for bootstrap's `input-group`. This input group includes the input (actually a div) and the button. The click event is on the input group in this case.

cheekw:
> I updated my above comment with checkboxes to track which tasks are complete. After reviewing the UI again, looks like most of the concerns have been fixed. The gene selector was changed but not fixed properly and 2 other things were not addressed (listed in above comment).
> 
> The issue (#135) describes how the app is supposed to behave when 1 of `Study` or `NCBIGene_ID` are present in the URL, when both are present, and neither are present.
> 
> No url parameters:
> 
> * https://staging.gudmap.org/deriva-webapps/plot/?config=study-violin
> * show both selectors active
> 
> Both `Study` and `NCBIGene_ID`:
> 
> * https://staging.gudmap.org/deriva-webapps/plot/?config=study-violin&Study=14-3PXT&NCBI_GeneID=1
> * show both selectors active
> * show `[Visualization with Gene & Study selectors](https://staging.gudmap.org/deriva-webapps/plot/?config=study-violin)` link in top right
> 
> Only `Study`:
> 
> * https://staging.gudmap.org/deriva-webapps/plot/?config=study-violin&Study=14-3PXT
> * hide study selector
> * show `[Visualization with Gene & Study selectors](https://staging.gudmap.org/deriva-webapps/plot/?config=study-violin)` link in top right
> 
> Only `NCBIGene_ID`:
> 
> * https://staging.gudmap.org/deriva-webapps/plot/?config=study-violin&&NCBI_GeneID=1
> * disable gene selector
> * show `[Visualization with Gene & Study selectors](https://staging.gudmap.org/deriva-webapps/plot/?config=study-violin)` link in top right

@jrchudy 
- behavior updated to be the same

cheekw:
added
- title in groupby object and used that instead to fix the issue with dropdown with display_patterns `xaxis.group_keys[x].title`

jrchudy:
2 issues still: 
1. The case when both `Study` and `NCBIGene_ID` are in the url, both selectors should be **visible AND active**. 
image
The older implementation looks like this:
image

2. The way gene selector is being disabled still makes the input look like it is enabled with a slightly darker grey background.
image
The older implementation looks more like this with a darker background and the button is not shown:
image

cheekw:
> 2 issues still:
> 
> 1. The case when both `Study` and `NCBIGene_ID` are in the url, both selectors should be **visible AND active**.
> 
> image
> 
> The older implementation looks like this:
> image
> 
> 2. The way gene selector is being disabled still makes the input look like it is enabled with a slightly darker grey background.
> 
> image
> 
> The older implementation looks more like this with a darker background and the button is not shown:
> image

Thanks for the catches.
- behavior logic: if all requestable selectors with href are on, they are now all active
- disabled styling is made to be consistent with the image shown above.

jrchudy (approved):

PR #147, opened 2023-05-24
React heatmap

KenyShah24:
Heatmap component in React js


src/components/heatmap/heatmap.tsx
```diff
@@ -0,0 +1,344 @@
...
+    const z = tuple.data[configData.zColumn];
+    const title = tuple.data[configData.titleColumn];
+    const id = tuple.data[configData.idColumn];
+    for (let i = 0; i < heatmaps.length; i++) {
+      if (Object(heatmaps[i]).title === title) {
+        Object(heatmaps)[i]['id'] = id;
+        hm = heatmaps[i];
+      }
+    }
+    if (hm === null) {
```
jrchudy, line 167: Can you add a comment explaining why this check is being done in conjunction with the above for loop?

src/components/heatmap/heatmap.tsx
```diff
@@ -0,0 +1,344 @@
...
+      hm = { 'title': title, 'rows': { y: [], x: [], z: [], type: 'heatmap' } };
+      heatmaps.push(hm);
+    }
+    let rowIndex = Object(hm).rows.y.indexOf(y);
+    if (rowIndex < 0) {
```
jrchudy, line 172: Why is this check being done to change rowIndex?

src/components/heatmap/heatmap.tsx
```diff
@@ -0,0 +1,344 @@
...
+  /**
+   * @param page : Page object fecthed from response of ermrestURI
+   */
+  const readAll = (page: any) => {
+    for (let i = 0; i < page.tuples.length; i++) {
+      addData(page.tuples[i]);
+    }
+    if (page.hasNext) {
+      Object(header).action = 'page';
+      page.next.read(1000, header).then(readAll).catch(function () {
```
jrchudy, line 138: same comment about changing the function syntax to use arrow syntax

src/components/heatmap/heatmap.tsx
```diff
@@ -0,0 +1,344 @@
...
+      });
+    } else {
+      setHeatmaps(() => (heatmaps));
+    }
+    setNCBIGeneID(() => (page.tuples[0].data.NCBI_GeneID));
+  }
+  /**
+   * 
+   * @param tuple : Tuple object retrieved from page
+   * Adds data(x,y,z values) to heatmap objects
```
jrchudy, line 150: Can you add more details about how this is adding the data?

For instance, is this adding 1 y and 1 x value if they aren't in the set yet? And then the corresponding z values?

src/components/heatmap/heatmap.tsx
```diff
@@ -0,0 +1,344 @@
...
+   * yaxis: {tickangle, tickvals, ticktext, tickfont} 
+   */
+
+  const createPlotLayoutParams = (plot: any = {}) => {
+    const configPresentation = Object.assign({}, config.presentation);
+    let layout = {};
+    if (plot) {
+      const longestXTick = plot?.rows?.x.reduce((a: any, b: any) => { return a?.length > b?.length ? a : b; });
+      const longestYTick = plot?.rows?.y.reduce((a: any, b: any) => { return a?.length > b?.length ? a : b; });
+      const inputParams = {
```
jrchudy, line 283: This needs a type defined for this object. Since we are creating the object ourself, we can define each property we know will be present and their types.

src/components/heatmap/heatmap.tsx
```diff
@@ -0,0 +1,344 @@
...
+     * 		r: right margin of the heatmap,
+     * 		b: bottom margin of the heatmap,
+     * 		l: left of the heatmap
+     * 	},
+     * 	xTickAngle: inclination of x axis labels,
+     *  yTickAngle: inclination of y axis labels,
+     * 	tickFont: font to be used in labels
+     * }
+     */
+  const getHeatmapLayoutParams = (input: any, longestXTick: number, longestYTick: number, lengthY: number) => {
```
jrchudy, line 206: `input` should have a type defined (see comment below).

The function should also have a return type defined since we are intending for this to always be called one way and always return a `layoutParams` object we are defining ourself.

src/components/heatmap/heatmap.tsx
```diff
@@ -0,0 +1,344 @@
...
+      yTickAngle = -90;
+      lMargin = 30;
+      rMargin = 20;
+    } else {
+      yTickAngle = 0;
+      lMargin = 20 + longestYTick * 7;
+      rMargin = 5;
+    }
+
+    const layoutParams = {
```
jrchudy, line 236: This should have a type as well. Similar comment to the one I made below about `inputParams`.

src/components/heatmap/heatmap.tsx
```diff
@@ -0,0 +1,344 @@
...
+        <a style={{ marginLeft: '1%' }}
+          href={`${window.origin}/chaise/recordset/${ConfigService.ERMrest.createPath('2', 'Gene_Expression', 'Array_Data', facet)}
+          ?pcid=${Object(header).cid}&ppid=${Object(header).pid}`} target='_blank' rel='noreferrer'>
+          View Array Data Table</a>
+      </ChaiseToolTip>
+      {/* Iterating over the heatmaps fetched for the given NCBIGeneID */}
+      {heatmaps.map((heatmap: any): JSX.Element => {
+        const { plot, layout } = createPlotLayoutParams(heatmap);
+        return <div id={`heatmap_${heatmap.id}`} key={`heatmap_${heatmap.id}`}>
+          <HeatmapPlot key={`heatmap_${heatmap.id}`} data={plot} layout={layout} />
```
jrchudy, line 337: Does this need a key too since the parent `<div>` that is being repeated has the same `key` value defined?

jrchudy (changes requested):
The functionality looks good, but we want to improve the typing as much as we can. Same with commenting bits of code that are doing something complicated or potentially confusing.

src/components/heatmap/heatmap.tsx
```diff
@@ -0,0 +1,378 @@
...
+}
+
+const Heatmap = ({
+  config,
+}: HeatmapProps): JSX.Element => {
+
+  // since we're using strict mode, the useEffect is getting called twice in dev mode
+  // this is to guard against it
+  const setupStarted = useRef<boolean>(false);
+  const [invalidConfigs, setInvalidConfigs] = useState<string>(['']);
```
jrchudy, line 54: this type should be `useState<string[]>(['']);

src/components/heatmap/heatmap.tsx
```diff
@@ -0,0 +1,378 @@
...
+  const [configErrorsPresent, setConfigErrorsPresent] = useState<boolean>(false);
+  const [heatmaps, setHeatmaps] = useState<any>([]);
```
jrchudy, line 56: this type should be `useState<any[]>([]);

jrchudy (changes requested):
Last 2 comments to fix then we can discuss about merging the code.

PR #149, opened 2023-05-30
React boolean search

PoornimaVKrishnan:


src/components/boolean-search/boolean-table.tsx
```diff
@@ -0,0 +1,307 @@
...
+            {location?.map((option: any, index: any) => {
+              return (
+                <option key={index}>
+                  {option.Pattern_Location}
+                </option>
+              );
+            })}
+          </select>
+        ),
+        actions: <button onClick={(e) => handleRemoveRow(e, index)}>Remove</button>
```
jrchudy, line 247: I don't think this "data"  row is being used. I don't see a button named "Remove" anywhere on the page. Where I would expect this to show, I see the contents of the `Cell` from the `columns` constant.

src/components/boolean-search/boolean-table.tsx
```diff
@@ -0,0 +1,307 @@
...
+
+    updatedRows[index] = updatedRow;
+    setRows(updatedRows);
+  };
+
+  const handleActiveRow = (e: any, index: any) => {
+    setActiveRow(index);
+  };
+
+  const columns: any = useMemo(
```
jrchudy, line 135: Please explain what the structure of this array of objects is and how it relates to the next constant (`data`) defined below.

There might be a type for it in react documentation that we can use instead of `any[]`, maybe something like `ReactTableColumn[]`.

src/components/boolean-search/boolean-table.tsx
```diff
@@ -0,0 +1,307 @@
...
+              onChange={(e) => handleChange(index, 'stageTo', e.target.value)}
+            >
+              {row.toStageOptions?.map((option: any, index: any) => {
+                return <option key={index}>{option.Name}</option>;
+              })}
+            </select>
+          </>
+        ),
+        pattern: (
+          <select
```
jrchudy, line 222: There should be an option in the select to set the pattern to empty string (which clears it from the query).

src/components/boolean-search/boolean-table.tsx
```diff
@@ -0,0 +1,307 @@
...
+            value={row.pattern || defaultValues.pattern || ''}
+            onChange={(e) => handleChange(index, 'pattern', e.target.value)}
+          >
+            <option style={{ display: 'none' }} disabled selected value=''></option>
+            {pattern?.map((option: any, index: any) => (
+              <option key={index}>{option.Pattern}</option>
+            ))}
+          </select>
+        ),
+        location: (
```
jrchudy, line 232: There should be an option in the select to set the location to empty string (which clears it from the query).

src/components/boolean-search/boolean-table.tsx
```diff
@@ -0,0 +1,307 @@
...
+              );
+            })}
+          </select>
+        ),
+        actions: <button onClick={(e) => handleRemoveRow(e, index)}>Remove</button>
+      })),
+    [rows, strength, pattern, location, stageFrom, source]
+  );
+
+  const tableInstance = useTable({
```
jrchudy, line 252: Please explain what the benefits of using this react library are.

It seems like a lot of extra information to learn/understand for something that can be defined ourself. The other concern is about support for this library. I don't expect it to change, but if something does that means more overhead for us to come back and fix this instead of it being our own `.tsx` implementation.

src/components/boolean-search/boolean-table.tsx
```diff
@@ -0,0 +1,307 @@
...
+  setRows,
+  strength,
+  pattern,
+  location,
+  stageFrom,
+  source,
+  changeFiltersDisplayText,
+  addClicked,
+  setAddClicked
+}: any) : JSX.Element=> {
```
jrchudy, line 17: We should define our own type for this component instead of using `any`. This example is how we have done it in most of our components in chaise too:
https://github.com/informatics-isi-edu/deriva-webapps/blob/master/src/components/search-input.tsx#L10-L38

src/components/boolean-search/boolean-table.tsx
```diff
@@ -0,0 +1,307 @@
...
+          })}
+        </tbody>
+      </table>
+    </div>
+  );
+};
+
+export default BooleanTable;
+
+BooleanTable.propTypes = {
```
jrchudy, line 295: After handling the above comment for adding a type to the component, this can be removed.

src/assets/scss/_booleansearch.scss
```diff
@@ -0,0 +1,151 @@
...
+                }
+                .main-search-input {
+                    width: 25vw;
+                }
+                .input-fullwidth {
+                    width: 45vw;
+                }
+            }
+            .form-container {
+                margin-top: 30px;
```
jrchudy, line 68: This should be 10px comparing with the angularJS implementation

jrchudy (changes requested):
Along with the inline comments (that I will summarize some of here), I have a few other comments that should be fixed too.

Bugs:
 - [ ] There is a bug where after selecting an option in treeview, the "activeRow" is updated to the last row. It should remain on the current row that was selected
 - [ ] When I click validate, only the last row shows a validation error message
     image

Presentation:
 - [x] the indentation between the Stages "from" dropdown and "To" text needs to be increased.
 - [x] empty string option for location and pattern dropdowns
 - [x] adjust the modal title and modal size
 - [x] adjust HTML structure of the `top-panel-container`
 - [x] create `chaise-input-group` for search input and buttons
 - [ ] create `chaise-btn-group` for buttons after search input
 - [ ] fix the styles of the `sidePanFiddler`

The other concern is about the `react-table` component. This looks good but I'm concerned about using an outside library for something we can do ourselves unless there is some functionality of this library that we are greatly benefitting from.

PoornimaVKrishnan:
Resolved all comments. Please review again

src/pages/boolean-search.tsx
```diff
@@ -0,0 +1,561 @@
...
+           <DisplayValue addClass value={{ value: 'Boolean Search', isHTML: true }} />
+        </h1>
+    </div>
+    <div className='bottom-panel-container'>
+      <div className={`resizable-panel resizable ${treeviewOpen ? 'open-panel' : 'close-panel'}`}>
+        <div className='treeview-panel'>
+          <TreeView />
+        </div>
+        <div className='resize-btn'>
+          <button
```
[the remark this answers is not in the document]
jrchudy: The rule is added, but the width is still be calculated differently (the old width was 21px and the new one is being set as 30px). The button is also detached from the line by a few pixels.
new:
image
old:
image

src/pages/boolean-search.tsx
```diff
@@ -0,0 +1,561 @@
...
+  }
+
+  if (!booleanSearchProps) {
+    return <ChaiseSpinner />;
+  }
+
+  return (<>
+  <Modal showModal={showModal} handleModalClose={handleModalClose}/>
+  <div className='boolean-container'>
+    <div className='title-container'>
```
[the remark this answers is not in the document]
jrchudy: Please add `app-content-container` and change `boolean-container` to `boolean-search-container`.

package.json
```diff
@@ -15,6 +15,7 @@
     "@types/plotly.js": "^2.12.18",
     "@types/plotly.js-basic-dist-min": "^2.12.1",
     "@types/react-plotly.js": "^2.6.0",
+    "@types/react-table": "^7.7.14",
```
jrchudy, line 18: Please remove `react-table` from here.

jrchudy (changes requested):
Changes look good. Thank you for rewriting the boolean-table component. Last couple of comments:
 - [ ] the spacing and size of `sidePanFiddler` is still off. See comment attached to previous comment from last review
 - [ ] remove `react-table` from package.json
 - [ ] clean up code for the `*-container` (fix class names and remove commented code from previous implementation)
 - [ ] fix indentation - some places in the code are harder to follow because of inconsistent indentation
   - In Visual Studio Code on Mac, `option + shift + f` will automatically fix indentation throughout the whole file you are looking at when the hotkey was pressed

src/components/boolean-search/boolean-table.tsx
```diff
@@ -0,0 +1,242 @@
...
+
+  }, [rows, activeRow]);
+
+  const handleRemoveRow = (evt: any, index: any) => {
+    const updatedRows = rows.slice(0, index).concat(rows.slice(index + 1));
+    setRows(updatedRows);
+    evt.stopPropagation();
+  };
+
+  const handleClearFilter = (e: any, index: number) => {
```
jrchudy, line 94: type for event

src/components/boolean-search/boolean-table.tsx
```diff
@@ -0,0 +1,242 @@
...
+    const updatedRows = [...rows];
+    updatedRows[index] = {
+      ...defaultValues
+    };
+    setRows(updatedRows);
+  };
+
+  const handleChange = (index: any, field: string, value: any) => {
```
jrchudy, line 102: type for index and value

src/components/boolean-search/boolean-table.tsx
```diff
@@ -0,0 +1,242 @@
...
+      updatedRow.stageTo = { ...updatedRow.stageTo, Name: value };
+    } else {
+      updatedRow[field] = value;
+    }
+
+    updatedRows[index] = updatedRow;
+    setRows(updatedRows);
+  };
+
+  const handleActiveRow = (e: any, index: number) => {
```
jrchudy, line 122: type for event

src/hooks/boolean-search.ts
```diff
@@ -0,0 +1,214 @@
...
+  return {
+    errors,
+    strengthData,
+    stageFromData,
+    patternData,
+    locationData
+  };
+};
+
+export const getSourceOptions = async (sources: any) => {
```
jrchudy, line 198: type for sources, `any[]`

src/pages/boolean-search.tsx
```diff
@@ -0,0 +1,554 @@
...
+   */
+  const togglePanel = () => {
+    setTreeviewOpen(!treeviewOpen);
+  };
+
+  /**
+   * Handles changing the search input field
+   * @param {any} event- change event
+   */
+  const changeSearch = (event: any) => {
```
jrchudy, line 104: type for event

src/pages/boolean-search.tsx
```diff
@@ -0,0 +1,554 @@
...
+    };
+    
+    const url = window.location.origin + '/chaise/recordset/' +
+    ConfigService.ERMrest.createPath('2', 'Gene_Expression', 'Specimen', null, customFacet)
++ '?pcid=' + headerInfo.cid + '&ppid=' + headerInfo.pid;
+    window.open(url, '_blank');
+    
+  }
+
+  const validateOtherParams = (invalidSource: any, submitQuery: boolean) => {
```
jrchudy, line 141: `any[]` type for invalidSource

src/pages/boolean-search.tsx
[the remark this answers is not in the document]
jrchudy: Looks good, thank you!

src/pages/boolean-search.tsx
```diff
@@ -0,0 +1,551 @@
...
+  }
+
+  if (!booleanSearchProps) {
+    return <ChaiseSpinner />;
+  }
+
+  return (<>
+  <Modal showModal={showModal} handleModalClose={handleModalClose}/>
+<div className='app-container'>
+  <div className='boolean-search-container'>
```
jrchudy, line 412: Add `app-content-container` to this div

jrchudy (approved):

jrchudy (changes requested):
2 bugs that I found after reviewing the functionality from the manual test cases document:
 - [x] when the page loads, the first row in boolean search should be "active" so the user can add a term from treeview without clicking a row first
 - [x] The document below has 4 test cases in it, following test case 2, I was able to determine that the validation isn't working properly when the user types directly into the box
   - follow "2. Querying using the textbox" from the test document and we can see that the 2nd bullet isn't happening.
     - when `invalid` values are entered in Strength, FromStage, ToStage, pattern, and location, theose fields should show validation errors
     - This must be what the `validateOtherParams` function was doing in the `angularJS` implementation that you asked me about. I didn't realize that a user could type into the text box directly and produce an invalid query
   - https://github.com/informatics-isi-edu/deriva-webapps/wiki/Test-Cases-for-Boolean-Search-application

jrchudy:
When I write this query into the box:
`strength{in "anatomical source" FromStage..ToStage pt=pattern lc=location} AND strength{in "anatomical source" FromStage..ToStage pt=pattern lc=location}`

It is being changed to:
`{in '' trength{in "anatomical source" FromStage..ToStage pt=pattern lc=location} AND {in '' trength{in "anatomical source" FromStage..ToStage pt=pattern lc=location}`

jrchudy:
The changes to fix the validation are working as we expect now. 2 issues that I found while messing around with the strength, To, and From dropdowns:
 - selecting `uncertain` in the `Strength` dropdown puts `uuncertain` into the query box. This should only be adding `u`
   - typing only `u` for strength then clicking `validate` passes the validation and selects `uncertain` properly, but the query in the search box is updated incorrectly to `uuncertain`
 - selecting `''` in `Strength` dropdown properly clears the strength value in the search box but selects `present` instead of `''` in the dropdown
   - removing the strength option from the search box and clicking `validate` does validation properly but the value shown in the strength dropdown does not match the displayed value (`present` is selected instead of `''`)
 - selecting `''` in `To` and `From` dropdown properly clears the to and from values in the search box but the actual selected string in the dropdown is `TS16` and `P110` by default
   - removing the `To` or `From option from the search box and clicking `validate` does validation properly, but the value shown in the To or From dropdown does no match the displayed value (`TS16` is selected instead of `''` for To. `P110` is selected instead of `''` for From.)

PR #151, opened 2023-06-05
Added heatmap plot type for plot app

KenyShah24:


src/hooks/chart.ts
```diff
@@ -346,6 +392,10 @@ const parsePlotData = (
       );
       additionalLayout = layout;
       return parseResult;
+    } else if (plot.plot_type === 'heatmap') {
+      const heatmapData = parseHeatmapResponse(currTrace, plot, responseData, additionalLayout);
```
jrchudy, line 396: No need to pass `additionalLayout` into this function since it's an empty object.

src/hooks/chart.ts
```diff
@@ -817,6 +882,160 @@ const parseBarResponse = (trace: Trace, plot: Plot, responseData: ResponseData)
...
+
+/**
+ * Parses response data obtained for every trace within the plot
+ *
+ * @param trace current trace the plot config
+ * @param plot plot config
+ * @param responseData data received from request to be parsed
+ * @returns
+ */
+const parseHeatmapResponse = (trace: Trace, plot: Plot, responseData: ResponseData, additionalLayout: any) => {
```
jrchudy, line 894: `additionalLayout` is not needed as a parameter. No contents should be defined on this object since this function is supposed to "create" that layout object and return it to `parsePlotData`.

src/hooks/chart.ts
```diff
@@ -817,6 +882,160 @@ const parseBarResponse = (trace: Trace, plot: Plot, responseData: ResponseData)
...
+    });
+    // Getting the longest x tick in the given data to determine margin and height values in getHeatmapLayoutParams function
+    const longestXTick = result.x.reduce((a: any, b: any) => { return a?.length > b?.length ? a : b; });
+    const longestYTick = result.y.reduce((a: any, b: any) => { return a?.length > b?.length ? a : b; });
+    const inputParams = {
+      width: typeof plotly?.layout.width !== 'undefined' ? plotly?.layout.width : 1200,
+      xTickAngle: typeof plotly?.layout.xaxis?.tickangle !== 'undefined' ? plotly?.layout.xaxis?.tickangle : 50,
+    }
+    layoutParams = getHeatmapLayoutParams(inputParams, longestXTick?.length, longestYTick?.length, result.y?.length);
+    additionalLayout = {
```
jrchudy, line 942: No need to use the previous `...additionalLayout`, `...additionalLayout.xaxis`, or `...additionalLayout.yaxis`, from the function parameter since we are creating this object here.

If we are defining our types properly, we shouldn't need to do `Object(layoutParams)`. Please explain why it needs to be done this way or remove it.

jrchudy (changes requested):
Some functionality issues that I noticed:
 - When there is a `yaxis.tick_display_markdown_pattern`, the graph gets squished:
    image
 - When there is a `xaxis.tick_display_pattern`, the bars get stretched out (which is different when there is no pattern):
    image
 - `graphic_link_pattern` needs to be an array of strings
 - How you are using additional layout is very confusing and hard to follow between the 3 functions

src/hooks/chart.ts
```diff
@@ -817,6 +882,160 @@ const parseBarResponse = (trace: Trace, plot: Plot, responseData: ResponseData)
...
+    y: [], // y data
+    z: [], // z data
+    text: [], // text data
+    legend_clickable_links: [], // array of links for when clicking legend
+    graphic_clickable_links: [], // array of links for when clicking graph
+  };
+  let layoutParams = {};
+  let yIndex = 0;
+  const { config, plotly } = plot;
+  const { xaxis, yaxis, format_data_x = false, format_data_y = false } = config; // why is format_data_x not in xaxis?
```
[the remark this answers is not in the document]
KenyShah24: So should I remove these all properties or just the format ones?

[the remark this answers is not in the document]
jrchudy: How you have this is good. I was responding to your comment to give a reason why it's different.

Ideally, what you said is what we should do but there's no need to update it right now.

src/hooks/chart.ts
```diff
@@ -817,6 +874,144 @@ const parseBarResponse = (trace: Trace, plot: Plot, responseData: ResponseData)
...
+        result.x.push(value.toString());
+      }
+    });
+
+    // Getting the longest x tick in the given data to determine margin and height values in getHeatmapLayoutParams function
+    const inputParams = {
+      width: typeof plotly?.layout.width !== 'undefined' ? plotly?.layout.width : 1200,
+      xTickAngle: typeof plotly?.layout.xaxis?.tickangle !== 'undefined' ? plotly?.layout.xaxis?.tickangle : 50,
+    }
+    layoutParams = getHeatmapLayoutParams(inputParams, longestXTick?.length, longestYTick?.length, result.y?.length);
```
jrchudy, line 941: `layoutParams` is being calculated for each `row` in `responseData`. Let's move this outside of `responseData.forEach(...)` so it is only done once after we create the `result` and calculate the longest x and y ticks.

jrchudy (changes requested):
Two things to address:
 - [ ] move `getHeatmapLayoutParams()` outside of the `forEach` loop
 - [ ] In my plot-config.js, I defined `plotly.layout.title` with NO `config.title_display_markdown_pattern`. The title I defined is not showing above the graph. 
    - I'm not sure why this is happening, but could you make sure that `plotly.layout` is being set as the "base" layout object, then any properties defined in config (like `config.title_display_markdown_pattern`) should override what is initially defined
    - something like:
       ```
       const layout = plotConfig.plotly.layout;
       
       if (plotConfig.config.title_display_markdown_pattern) {
         let tempTitle = renderHandlerbarsTemplate(config.title_display_markdown_pattern, templateParams);
         tempTitle = renderMarkdown(tempTitle);
         layout.title = tempTitle;
       }
       ```

jrchudy (approved):

Issue #74, opened 2019-08-20
heatmap: better handle the 404 error


hongsudt:
https://dev.gudmap.org/deriva-webapps/heatmap/#2/Gene_Expression:Array_Data_view/NCBI_GeneID=1
Throws a very weird error:
```
Configuration Errors
Change the following in the heatmap-config.js file:    
Error while reading data from Ermrestjs
```
In this case, it seems like NCBI_GeneID=1 doesn’t exist in the table.. 

We should handle this error better.. 

Note: 
- GeneID=17740 will work fine. 
- Use record app to check whether the data exist: https://www.gudmap.org/chaise/record/#2/Gene_Expression:Array_Data_view/NCBI_GeneID=1

vipul-21:
On clicking the above mentioned link. I am getting the following error:
1. When not logged in.
<img width="1440" alt="Screen Shot 2019-08-21 at 11 06 39 AM" src="https://user-images.githubusercontent.com/8600441/63456602-24556400-c404-11e9-979e-810ca9202ba2.png">

2. When logged in.
<img width="1440" alt="Screen Shot 2019-08-21 at 11 07 40 AM" src="https://user-images.githubusercontent.com/8600441/63456633-3b945180-c404-11e9-8755-3d20835fb91f.png">

There was no `Configuration Error`. @hongsudt Can you please let me know how to replicate the issue if it still exists.

jrchudy:
The issue is with the deriva-webapps link above not the chaise link.

https://dev.gudmap.org/deriva-webapps/heatmap/#2/Gene_Expression:Array_Data_view/NCBI_GeneID=1

When navigating to this link, make sure your ad blocker is turned off for that domain so that you don't get an angular error and can instead see the DB error.

vipul-21:
There are 2 cases to handle for 404 Not Found:
1. There is no heatmap data available for the given “id” (but the record with that id exists)
2. There is no record with the given “id” (and therefore there is no heatmap data that exists)

- [ ] Need to find out(in the code base)  if we actually check for the record existence in heatmap, if that's the case then the error for both the cases can be different.
For 1. the error will be "There is no heatmap data for the given 'id'"
For 2. the error will be "No record for the given 'id'"

If the above is not true, then we need to display the same error for both the cases "There is no heatmap data for the given 'id'"

PR #152, opened 2023-06-07
Added error message for invalid id and no data scenario

KenyShah24:
2 cases to handle for 404 Not Found:

1) There is no heatmap data available for the given “id” (but the record with that id exists)
2) There is no record with the given “id” (and therefore there is no heatmap data that exists)

Issue link : https://github.com/informatics-isi-edu/deriva-webapps/issues/74

### DIFF
--- a/src/components/heatmap/heatmap.tsx
+++ b/src/components/heatmap/heatmap.tsx
@@ -2,6 +2,7 @@ import '@isrd-isi-edu/deriva-webapps/src/assets/scss/_heatmap.scss';
 
 // hooks
 import { useEffect, useRef, useState } from 'react';
+import useError from '@isrd-isi-edu/chaise/src/hooks/error';
 
 // utilties
 import { chaiseURItoErmrestURI, getQueryParams, getCatalogId } from '@isrd-isi-edu/chaise/src/utils/uri-utils';
@@ -14,6 +15,8 @@ import ChaiseToolTip from '@isrd-isi-edu/chaise/src/components/tooltip';
 import { windowRef } from '@isrd-isi-edu/deriva-webapps/src/utils/window-ref';
 import HeatmapPlot from '@isrd-isi-edu/deriva-webapps/src/components/heatmap/heatmap-plot';
 import { HeatmapConfig } from '@isrd-isi-edu/deriva-webapps/src/models/heatmap-config';
+import { ChaiseError } from '@isrd-isi-edu/chaise/src/models/errors';
+import ChaiseSpinner from '@isrd-isi-edu/chaise/src/components/spinner';
 
 
 export type HeatmapProps = {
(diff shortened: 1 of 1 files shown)


src/components/heatmap/heatmap.tsx
```diff
@@ -84,32 +91,48 @@ const Heatmap = ({
...
+                if (pcid || ppid) {
+                  let tempHeader = { ...header }
+                  if (pcid)
+                    tempHeader.pcid = pcid
+                  if (ppid)
+                    tempHeader.ppid = ppid
+                  setHeader(() => (tempHeader));
+                }
+                ref.read(1000, header).then((page: any) => {
+                  readAll(page);
```
jrchudy, line 119: Before calling `readAll`, we should do the check here to see if there is any "array data". `ref.read()` is making a get request to the `ermrestUri` path which is to the array data table. 

`Page` is an API that Aref and I designed so you can safely call `Page.length` to check how many rows are returned for the `ermrestUri` request.

```
if (page.length === 0) {
  throw new ChaiseError('Data not available', 'No heatmap data available for the given ID<br>');
}

readAll(page);
```

Make sure to test that this error is being thrown/set properly. You will have to make a change to the `catch()` block right after this to make sure this error is shown. I think the catch will ignore any thrown error and just blindly `setInvalidConfigs`.

Maybe something like:
```
}).catch((err: any) => {
  if (err) {
    throw err;
  } else {
    setInvalidConfigs(['Error while reading data from Ermrestjs']);
    setConfigErrorsPresent(true);
  }
});
```

src/components/heatmap/heatmap.tsx
```diff
@@ -84,32 +91,48 @@ const Heatmap = ({
...
-        const pcid = getQueryParams(location.href).pcid
-        const ppid = getQueryParams(location.href).ppid
-        if (pcid || ppid){
-          let tempHeader = {...header}
-          if(pcid)
-            tempHeader.pcid=pcid
-          if(ppid)
-            tempHeader.ppid=ppid
-          setHeader(() => (tempHeader));
+    const geneId = location.hash.split('/')[2];
```
jrchudy, line 94: Add a comment explaining that the `location.hash` is expected to be in the form of `<catalog-id>/<schema>:<table>/NCBI_GeneID=<value>`. Also that this works specifically for genes and their associated data in the  `Array_Data_view`.

src/components/heatmap/heatmap.tsx
```diff
@@ -56,9 +59,13 @@ const Heatmap = ({
   const [invalidConfigs, setInvalidConfigs] = useState<string[]>(['']);
   const [configErrorsPresent, setConfigErrorsPresent] = useState<boolean>(false);
   const [heatmaps, setHeatmaps] = useState<any[]>([]);
+  const [noData, setNoData] = useState<boolean>(false);
+  const [isLoading, setIsLoading] = useState<boolean>(false);
```
jrchudy, line 63: Initialize `isLoading` to `true` since the app is "loading" initially until we know all the requests have returned.

src/components/heatmap/heatmap.tsx
```diff
@@ -84,32 +91,48 @@ const Heatmap = ({
...
-        ref.read(1000, header).then((page: any) => {
-          readAll(page);
-        }).catch(() => {
-          setInvalidConfigs(['Error while reading data from Ermrestjs']);
-          setConfigErrorsPresent(true);
-        });
-      }
-    });
+      }).catch((err: any) => {
+        setNoData(() => true);
```
jrchudy, line 133: We should do `setIsLoading(false)` in this catch block before throwing an error since we know the app is done setting up.

jrchudy (changes requested):
Thank you for going through and figuring this out. There are a few review comments to clean up the code

src/components/heatmap/heatmap.tsx
KenyShah24: I still need something to know that data is not there so the page return empty fragment

jrchudy (approved):

PR #155, opened 2023-06-15
Support hover template for heatmap type for plot app

KenyShah24:
Initially, when you hover over a section of the graph, the x/y axis values are displayed as default. However, we've enhanced the plot's capabilities to offer customizable hover text on the graph for all the plots. For example, now when you hover over a block of the heatmap, the hover text will appear based on the template defined as hovertemplate_display_pattern.
- NOTE: 
    - For the `violin` plot, the hover_template_displaypattern will only get applied to the scatter plot inside the violin plot(i.e. on hovering the datapoints) and not on the violin plot and the box plot. 
    - If one value or all values mentioned in the  `hovertemplate_display_pattern` is missing then also it will show the custom hover text the values that are available. For instance, if the `hovertemplate_display_pattern` is `Gene ID: {{{$url_parameters.Gene.data.NCBI_Ge}}}` where `NCBI_Ge` is not a valid key then it will just display `Gene ID` on hover.


config/plot-config-sample.js
```diff
@@ -567,6 +577,7 @@ var plotConfigs = {
         },
         traces: [
           {
+            hovertemplate_display_pattern: "Creation Date: {{{$row.RCT}}}<br>Vertical",
```
jrchudy, line 580: For consistency, this example should use 'Horizontal' since that's the orientation of the chart

jrchudy (changes requested):
Code changes look good. I found the following bugs:
 - For `study-violin`, all studies are being selected by default on page load now. This is supposed to be selecting the first study in the set
   - compare https://dev.gudmap.org/~jchudy/deriva-webapps/plot/?config=study-violin
   - with https://dev.gudmap.org/deriva-webapps/plot/?config=study-violin
 - For `gudmap-data-summary-responsive`, the scale of the xaxis is set to linear even though my configuration is set to use `log` (the sample uses `log` too)
   - compare https://dev.gudmap.org/~jchudy/deriva-webapps/plot/?config=gudmap-data-summary-responsive
   - with https://dev.gudmap.org/deriva-webapps/plot/?config=gudmap-data-summary-responsive
 - For `heatmap`, the `hovertemplate_display_pattern` is not showing based on the example in sample config
   - here in my user directory https://dev.gudmap.org/~jchudy/deriva-webapps/plot/?config=heatmap&NCBI_GeneID=11669
 - Also for `heatmap`, the default hoverinfo when no `hovertemplate_display_pattern` is defined shows "false" for x and y values instead of the axis labels
   - image

src/hooks/chart.ts
```diff
@@ -1090,4 +1127,113 @@ const updateWithTraceColData = (
...
+ */
+const defaultHoverTemplateDisplay = (
+  result: any,
+): void => {
+  const tempText: any[]=[];
+  if (result.data[0].type==='heatmap') {
+    result.data[0].hoverinfo= 'text';
+    result.data[0].z.forEach((zArr: any[],index: number)=>{
+      tempText.push([]);
+      zArr.forEach((val: any,i: number)=>{
```
jrchudy, line 1209: is `val` always a string or can it be a different type?

src/hooks/chart.ts
```diff
@@ -1090,4 +1127,113 @@ const updateWithTraceColData = (
...
+ * @param result trace object to be updated
+ */
+const defaultHoverTemplateDisplay = (
+  result: any,
+): void => {
+  const tempText: any[]=[];
+  if (result.data[0].type==='heatmap') {
+    result.data[0].hoverinfo= 'text';
+    result.data[0].z.forEach((zArr: any[],index: number)=>{
+      tempText.push([]);
```
jrchudy, line 1208: To avoid having multiple types (or using `any[]`), this whole block of code can be changed to:
```
const textArray = [];
zArr.forEach((val: any,i: number)=>{
  textArray.push(...)
})
tempText.push(textArray);
result.data[0].text=tempText;
```

src/hooks/chart.ts
```diff
@@ -1090,4 +1127,113 @@ const updateWithTraceColData = (
...
+  }
+}
+
+/**
+ * Sets the hovertext array and hoverinfo for all plots when hovertemplate_display_pattern is configured
+ * @param result data object to be updated
+ * @param textArray array of hover text
+ * @param trace from plot configs
+ */
+const setHoverText = (result: any, textArray: any, trace:Trace) => {
```
jrchudy, line 1233: Same comment about type for this array.

jrchudy (changes requested):
Functionality looks good. The code needs to have types cleaned up. I also noticed you are using `_renderHandlebarsTemplate`. 
 - [x] Use `ERMrest.renderHandlebarsTemplate` since functions beginning with `_` are intended to be private functions only used in ermrestJS.

I left a comment about `alertFunctions` that also applies to `templateParams`. It seems wrong to be passing both of these objects into every function that use them instead of having them available in the scope they are called in.
  - #### Edit: After discussing this in more details, the functions should be moved inside the hook so they have access to the alerts provider variables and the state variable defined for templateParams
  - [ ] move functions and refactor to remove unnecessary params that can be state variables

There is a bug introduced that isn't properly rendering x axis tick labels markdown patterns. 
  - [x] Compare with [https://dev.gudmap.org/deriva-webapps/plot/?config=study-violin](master) to see what is happening for this configuration and make sure it displays properly


Please update `plot-functionality.md` to include a section called `Graph Hover Events` after `Graph Click`. This should outline:
 1.  the functionality that was added to support hover template patterns. 
 2.  details about the "default" hover templates that our code is applying when not template is defined 
 3. drawbacks of functionality (violin plot case)
 4. what happens when 1 value is null in the template
    - note how this differs from other patterns since other patterns return null rather than ignoring the missing/null values

src/hooks/chart.ts
```diff
@@ -1090,4 +1127,113 @@ const updateWithTraceColData = (
...
+  trace: Trace, 
+  item: any, 
+  textArray: any[], 
+  alertFunctions?: any,
+  index: number=0, 
+  templateParams?: PlotTemplateParams,
+  ) => {
+  const hovertemplate_display_pattern = trace.hovertemplate_display_pattern; // use trace info
+  if(hovertemplate_display_pattern){
+    const validLink=ConfigService.ERMrest._renderHandlebarsTemplate(hovertemplate_display_pattern, { $self: { data: item },
```
jrchudy, line 1154: change to `renderHandlebarsTemplate`. Functions beginning with `_` are intended to be private and only used in ermrestJS.

src/hooks/chart.ts
```diff
@@ -327,12 +335,11 @@ export const useChartData = (plot: Plot) => {
   // Parse data on state changes to data or selectData
   useEffect(() => {
     if (data && !isDataLoading && !isInitLoading && !isFetchSelected) {
-      const parsedPlotData = parsePlotData(plot, data, selectData, templateParams);
+      const parsedPlotData = parsePlotData(plot, data, selectData, templateParams, alertFunctions);
```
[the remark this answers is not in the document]
KenyShah24: Yes, that's correct

src/hooks/chart.ts
```diff
@@ -809,463 +795,621 @@ const parseViolinResponse = (
...
+   */
+  const defaultHoverTemplateDisplay = (
+    result: any,
+  ): void => {
+    const tempText: string[][] & string[] = [];
+    if (result.data[0].type === 'heatmap') {
+      result.data[0].hoverinfo = 'text';
+      result.data[0].z.forEach((zArr: string[], index: number) => {
+        const textArr: string[]=[];
+        zArr.forEach((val: any,i: number)=>{
```
jrchudy, line 1024: I think this param type should be `val: string`

src/hooks/chart.ts
```diff
@@ -809,463 +795,621 @@ const parseViolinResponse = (
...
 
-  return result;
-};
+  /**
+   * Sets the hovertext array and hoverinfo for all plots when hovertemplate_display_pattern is configured
+   * @param result data object to be updated
+   * @param textArray array of hover text
+   * @param trace from plot configs
+   */
+  const setHoverText = (result: any, textArray: any, trace: Trace) => {
```
jrchudy, line 1050: I think this param type should be `textArray: string[]`

src/components/plot/chart-with-effect.tsx
```diff
@@ -146,33 +150,35 @@ const ChartWithEffect = ({ config }: ChartWithEffectProps): JSX.Element => {
...
-        {isParseLoading || isFetchSelected ? (
-          <ChaiseSpinner />
-        ) : (
-          <PlotlyChart
-            className='plotly-chart'
-            style={dynamicStyles}
-            ref={plotlyRef}
-            {...parsedData}
-            useResizeHandler
+    <AlertsProvider>
```
jrchudy, line 153: `AlertsProvider` doesn't need to be added here. Since we are using `AppWrapper` (and setting `includeAlerts`), the `AlertsProvider` will be automatically included

src/hooks/chart.ts
```diff
@@ -809,463 +795,621 @@ const parseViolinResponse = (
...
+   * @param textArray array for hover text
+   * @param index provided for heatmap case for corresponding y axis values
+   * @param templateParams provided for heatmap to access url_parameters
+   * @returns 
+   */
+  const updateHoverTemplateData = (
+    result: any,
+    trace: Trace,
+    item: any,
+    textArray: string[] & string[][],
```
jrchudy, line 965: Should this be `string[] & string[][]` or `string[] | string[][]`?

I'm not sure what the difference is if you could comment why one is preferred over the other. 

I think the case that might be `string[][]` is for heatmap plot type since the `text` array matches the `z` array?

jrchudy (changes requested):
One issue I noticed this time that I didn't catch in my last review was duplicating the `AlertsProvider`. Otherwise, the other changes are some minor type issues to address and then we can merge these changes.

I will do one more final review and test all plot types after those changes are done before I merge this into master.

jrchudy (approved):

Issue #162, opened 2023-06-28
Allow for plot data to come from a file


jrchudy:
We want to be able to fetch plot data from a file instead of ermrest. The use case mentioned said the file content would be CSV. We want to be able to fetch data from any API (not just ermrest) as json or csv, as well as from a file from any url that is json or csv. The following needs to be done to support this:
 - change `queryPattern` to be `url_pattern`
   - determine if there is a host or domain name in the url_pattern (dev.gudmap.org)
   - if there is no host/domain name, assume same host
 - add another property to `traces` to communicate the expected format the data is in
   - lets call this `response_format` (similar to the accept header)
     - this property should accept 'csv' | 'json' only
     - **what content types in `response.headers.content-type` do we want to match against?**
   - this property is optional
   - if not defined, try to figure out the type of the content returned (json or csv)
     - if we can't figure out type, assume json
     - **if something goes wrong because we assumed json, throw error? warning? or show "No Data"?**
 - Create a `.json` file and `.csv` file in your user directory on dev.gudmap
   - pick a plot config to test this for and copy the contents of the response into a `.json` file
   - to test loading data from ermrest as csv, append `?accept=csv` as a query parameter to the `queryPattern` (now url_pattern)
     - copy these contents into a `.csv` file
 - make changes in code to parse csv data properly into a json structure similar to what we've been working with from ermrest

KenyShah24:
For parsing CSV file we can either use the library or internal parser written within the code:

1. fast-csv: [@fast-csv/format] and [@fast-csv/parse] => We need only parser 
 **- Slow as compared to other parsers**
 - size: 8.75kb
3. react-papaparse:  easy to use, just Papa.parse(input, config), very fast, Can parse strings, local files, and even download remote files
, Fast mode for non-quoted data
, Streaming support for parsing large files, Automatic type conversion

, compliant and correctly handles line-breaks and quotations

 **- Has many such functionalities which we don't need causing an unwanted increase in the package size**
 - size: 71kb
4. csv-parse: Flexible with lot of options, works on big datasets,
 **- size 1.38MB**
5. csv-parser: lightest parser
 - size: 1.5kb
If we are only gonna have specific types of csv and we will always know it's format it's better to use the parser created in the code itself. That will be simpler. But if the csv files are out of our control and has many edge cases like having no header, all or some fields wrapped in quotes, linebreaks in quotes,etc then using library will be better.

KenyShah24:
[...]

PR #173, opened 2023-08-07
Plot data from file

KenyShah24:
This functionality enables to fetch data for the plot app from a specified file, which can be either in 'csv' or 'json' format. The data will be parsed and presented as a plot based on the content type of the file. To utilize this feature, two parameters can be configured in the config file:

- url_pattern: The URL from which the data will be fetched.
- response_format (optional): It specifies the file type to be used with the url_pattern parameter.

Note: The test data for this issue can be found on https://github.com/informatics-isi-edu/deriva-ui-test-data repository
 Resolves: https://github.com/informatics-isi-edu/deriva-webapps/issues/162


src/models/plot.ts
```diff
@@ -105,7 +106,9 @@ export type TraceConfig = {
   y_col?: string[];
   z_col?: string[];
   legendwidth?: number;
-  queryPattern?: string;
+  url_pattern?: string;
+  queryPattern?:string,
```
jrchudy, line 110: Add a comment saying this will be deprecated

src/models/plot.ts
```diff
@@ -126,3 +129,21 @@ export const plotAreaFraction = 0.95;
 //NOTE: Consider moving this threshold to the useWindowSize hook if it will be utilized by other webapps.
 export const screenWidthThreshold = 1000;
 
+//When no domain is provided in url_pattern in traces then consider this as default domain to fetch data
+export const defaultDomain = '/dev.gudmap.org';
```
jrchudy, line 133: domain should not include `/` at the start of it.

src/models/plot.ts
```diff
@@ -126,3 +129,21 @@ export const plotAreaFraction = 0.95;
...
+
+//Alert messages
+export const invalidKeyAlert='Invalid key provided for hover template display pattern!';
```
jrchudy, line 136: Similar to what we did in chaise, can you move these alert messages to a separate file called `message-map.ts`?

The file is in `src/utils/message-map.ts` in chaise.

src/utils/string.ts
```diff
@@ -267,4 +268,45 @@ export const wrapText = (text: string, width: number, wrapLimit: number) => {
...
+
+/**
+ * 
+ * @param data It can be either csv or json data
+ * @returns true if data is of json type and false for other types
+ */
+export const isDataJSON = (data: any) => {
+  try {
+    const parsedData=JSON.parse(JSON.stringify(data));
+    console.log('data ',parsedData,typeof parsedData);
```
jrchudy, line 281: Remove this console.log

src/utils/string.ts
```diff
@@ -1,6 +1,7 @@
 import { ConfigService } from '@isrd-isi-edu/chaise/src/services/config';
 import { windowRef } from '@isrd-isi-edu/deriva-webapps/src/utils/window-ref';
-import { dataFormats } from '@isrd-isi-edu/chaise/src/utils/constants';
+import axios from 'axios';
```
jrchudy, line 3: remove this unused import

src/hooks/chart.ts
```diff
@@ -349,7 +362,11 @@ export const useChartData = (plot: Plot) => {
       // request for each trace
       plot.traces.map((trace) => {
         // Check for queryPattern(dynamic link) parameter in traces, if not defined then check for uri(static link)
-        if (trace.queryPattern) {
+        if (trace.url_pattern) {
+          const url = createUrlFromPattern(trace.url_pattern);
+          return axios.get(url);
```
jrchudy, line 367: Can you comment why axios is being used here? Why aren't we using `ConfigService.http.get`?

If you look at `src/utils/config.ts` in chaise, you can see that `http` is defined like this:
```
static get http() {
  return ConfigService._setupDone && ConfigService._server ? ConfigService._server.http : axios;
}
```

and will default to axios if needed

src/hooks/chart.ts
```diff
@@ -1432,42 +1461,93 @@ export const useChartData = (plot: Plot) => {
...
-        plotlyData.push(heatmapData.result);
+      //If the response_format is configured then check the format against type of file and parse the data accordingly
+      if (responseData && currTrace.response_format) {
+        //If the given format is not from the allowed types then show an alert warning
+        if (!(validFileTypes.includes(currTrace.response_format)) && !alertFunctions.alerts.some(
+          (alert) => alert.message.includes(invalidResponseFormatAlert))) {
+          alertFunctions.addAlert(alertMsg + invalidResponseFormatAlert, ChaiseAlertType.WARNING);
+        }
+        //If the given format is csv and the content of file is also of type csv then parse the data using csv parser
+        else if (currTrace.response_format === 'csv' && !isDataJSON(responseData)) {
```
jrchudy, line 1481: `isDataJson(responseData)` is being called 4 times in this function for different conditionals. This should be done once outside of the `if (responseData && currTrace.response_format)` conditional since the check won't change. Then it can be reused in all 4 places.

jrchudy (changes requested):
The changes in the code look good. There are some places that I asked you to add comments . The other issues I noticed are the following:
 - [x] When I use `url_pattern`, study violin does not load the data properly. This seems to work for other data types
   - once this is fixed, I will test fetching data from a file for each plot type
 - [x] The comments I made in standup about the regression testing document
   - a separate `plot-config-test.js` should be created in dev-docs that is used for testing purposes
   - there should be at least 2 bar plot configurations that detail a "simple" bar plot with no clickable links
   - the 2nd should have a markdown pattern for the title, each axis title, the non numeric axis, the legend, and the graphic link
   - all of this should be detailed in the document to communicate that each of these properties needs to be tested and navigate to the "right" expected page
   - Do this for each plot type that we support (pie, scatter, histogram, heatmap)
   - The links in the document should point to what is the expected path "after" uploading the `plot-config-test.js` to the "testing user's directory"
 - [ ] When I pulled your branch, I couldn't run make dist because the `package-lock.json` was out of sync with `package.json`

jrchudy (approved):

Issue #172, opened 2023-08-02
Add a selector for data for any plot type


jrchudy:
This issue is meant to be an incremental change that gets our functionality closer to what is described in issue #156.

For now, only 1 selector will be allowed to be added. This selector will be displayed as a dropdown with the data coming from an ermrest request or hard coded as part of the selector. 

### Configuration
```
plots:[{
  plotly: {
    layout: {...},
    config: {...}
  },
  config: {...},
  dropdown: {
    uid*:                   string - key for referencing the dropdown in other configuration properties,
    label*:                 string - name to display next to the dropdown,
    url_param_key?:         string - parameter name from url params that is associated with the dropdown,
    request_info: {
      url_pattern?:            string - string that allows for handlebars templating for fetching data,
      data?:                   array of objects - values to use in the dropdown if no url_pattern is provided, 
      default_value?:          string - the initial value to use on page load (url_param_key can be set to override this),
      value_key*:              string - column to use for templating in queries,
      selected_value_pattern?: string - a pattern string to show in the dropdown for each selected option
      tick_markdown_pattern*:  string - a markdown pattern to be used for the axis labels
    }
  },
  traces: [{...}]
}, ...]
```

#### Notes:
 - `url_param_key` can be defined to allow for a url parameter to be used to initialize the dropdown and fetch data using the url param value 
 - 1 of `url_pattern` or `data` has to be defined for this dropdown to have data in it
   - show configuration warning alert and show the dropdown as empty
   - still try to fetch data as if no option was selected and let the rest of the code handle any errors that occur
   - this property is used to fetch the value from the url and store it in `$control_values.<dropdown.uid>.values.<dropdown.value_key>`
 - `default_value` _should_ be set if fetching data for the plot relies on a value being selected
   - needs to be a valid value for `value_key` in each row that is returned from `url_pattern` or `data`
 - `selected_value_pattern` is optional. If not defined, use rowname (if url_pattern) or the `value_key`
 - This property will be accessed in templating user the `$control_values` namespace
   - this replaces `$url_parameters` that was introduced with violin plot
   - support both, mark `url_parameters` as deprecated
   - the selected data will be made available in templating
     - `$control_values.<dropdown.uid>.values.<dropdown.value_key>`

### Example with "data"
This example is for changing the consortium value that is present in the url_pattern.

```
plots: [{
  plot_type: 'bar',
  plotly: {...},
  dropdown: {
    uid: 'consortium',
[...]

Issue #161, opened 2023-06-28
Configurable User Controls - Grid Layout 


jrchudy:
To begin implementing configurable selectors for plot app, we want to change the way the layout of the app is being produced. Currently selectors are hardcoded to be above the specific plots that use them. This layout will allow for data modelers to configure which selectors to show and how they will interact with the data.

This task's focus is on the `grid_layout_config` and `layout` properties being defined in issue #156. `grid_layout_config` is an object that we are using to configure the "grid" component that we will use from [ReactGridLayout library](https://github.com/react-grid-layout/react-grid-layout). We want to support layout within a plot (`plots[0].grid_layout_config`) first, we will extend this to support a "global layout" that allows for both selectors and plots to be configured in the grid.

Steps to take for local layout:
 - install `ReactGridLayout` npm library
 - when `grid_layout_config` is defined, create a `ReactGridLayout` component above or below the plot based on configuration
 - translate the configuration language in `grid_layout_config` to the language that `ReactGridLayout` expects
   - our language uses snake case with `_` separating content, `ReactGridLayout` uses camel case
   - the language defined in issue #156 will be the set we allow data modelers to configure, there might be more properties we configure internally so please take a look at the properties allowed:
     - https://github.com/react-grid-layout/react-grid-layout#grid-layout-props
 - Once the grid is initialized with all the configured values, make sure you can add a selector into the grid
 
Steps to take for global layout:
  - Once the above has been done for the grid in a local layout, we need to also allow for the grid to be defined at the global level  to be able to display global selectors in a grid alongside the plots
  - an example to get working first would be to have 2 plots side by side

jrchudy:
We want to list the properties in the `grid_layout_config`, each control in `user_controls`, and each layout in `layout` that are required to be defined. We should be validating this list in the code and throwing proper errors when those properties are not defined. 

List of properties to validate:
 - `grid_layout_config`:
   - `cols` - might not need to be validated since we might have a default to use? 12 cols as default maybe?
     - fails validation for this control if not defined
       - **what does a failed validation mean?**
   - _please fill in more..._
 - `user_control`:
   - `uid` - required or can't map control to `$control_values` used for templating or map to an object in `layout`
     - fails validation for this control if not defined
   - `type` - required unless we want to assume a default is dropdown
     - we should probably enforce this as a requirement since other properties being defined depend on this being set
     - fails validation for this control if not defined
       - **what does a failed validation mean?**
[...]

Issue #156, opened 2023-06-15
Plot app: Configurable User Controls functionality in plot-config


jrchudy:
Currently only the violin plot has selectors that show above the plot that are hardcoded. If violin plot wants to be used for a different use case, some of these user controls might not make sense. We also have a use case to add a user control specifically for the `heatmap` plot type that allows for the column that is used for aggregation to be changed so the plot can show different "z values" for the same set of x/y data (issue #144 ).

The following properties are how we will add support for configurable user controls in plot-config:
```
layout: [
  { layout },
],
user_controls: [
  { control }
],
grid_layout_config: {},
plots: [{
  uid: 'plot-id',
  plotly: {
    layout: { ... },
    config: { ... }
  },
  config: {
    xaxis: {
      type_pattern: '...'
    },
    ... 
  },
  user_controls: [
    { control },
    { control },
    ...
  ],
  layout*: [
      { layout },
      { layout },
      ...
  ]
  grid_layout_config: {
    width?:              number - width in pixels, not needed if using responsive grid layout component,
    auto_size?:          boolean - height grows/shrinks to fit content,
    position?:           string - where the grid of selectors will show ('top' | 'bottom')
    breakpoints?:        object - map to identify when a different layout configuration should be used based on grid size,
    cols?:               string | object - number of columns to show, object used for breakpoints,
    margin?:             array of numbers | object - margin between grid components in pixels,
    container_padding?:  array of numbers | object - padding inside of each grid component in pixels,
    row_height?:         number - height of each row in pixels,
    is_draggable?:       boolean - controls ability to move all components,
    is_resizable?:       boolean - controls ability to resize all components
  },
  traces: [{
    url_pattern: '...',
    x_col_pattern: [ '...' ],
    y_col_pattern: [ '...' ],
    z_col_pattern: [ '...' ]
  }, ... ]
}]
```

Each individual `control` will be configured with the following:
```
* = required
? = optional

user_controls: [{
  uid*:                   string - key for referencing this user control; in other configuration properties,
  label*:                 string - name to display next to this user control,
  type*:                  string - the type of control to display, more info below about control types,
  url_param_key?:         string | object - parameter name or object of parameter names from url params that ire associated with this user control,
  request_info?: {
    url_pattern?:            string - string that allows for handlebars templating for fetching data for this user control,
    data?:                   array of objects - values to use in the user control if no query_pattern is provided, 
[...]

PR #177, opened 2023-09-12
Plot App: Added selector using React grid layout

KenyShah24:
This Pull request consists of resolving following 2 issues:
1. Configuring selector from config file to fetch data:
  - Currently, only one selector can be added, which will be presented as a dropdown menu. The data for this selector can either be sourced from a data array provided in the config file of plot app.
2. Configuring grid layout for selector from config file to decide the placement of the selector:
  - The primary objective of this task revolves around defining the properties related to grid layout configuration, as specified in issue #156. The grid_layout_config is an object designed to configure the "grid" component sourced from the ReactGridLayout library. Currently, our focus is on enabling layout customization within an individual plot (specifically, plots[0].grid_layout_config). Subsequently, we aim to expand this functionality to encompass a broader "global layout," enabling configuration of both selectors and plots within the grid as an enhancement.

Resolves :  https://github.com/informatics-isi-edu/deriva-webapps/issues/172
In Progress : https://github.com/informatics-isi-edu/deriva-webapps/issues/161

### DIFF
--- a/docs/user-docs/plot-app.md
+++ b/docs/user-docs/plot-app.md
@@ -17,8 +17,8 @@ Each object in the `plots` array can have the following parameters:
 1. `plot_title`: Title to be displayed above the line plot. DEPRECATED, use `plotly.layout.title` instead
 2. `x_axis_label`: Label to be displayed along the x-axis. DEPRECATED, use `plotly.layout.xaxis.title` instead
 3. `y_axis_label`: Label to be displayed along the y-axis. DEPRECATED, use `plotly.layout.yaxis.title` instead
-4. `plot_type`: The type of plot to be displayed i.e `line`, `bar`, `dot`, `area`, `dot-lines`, `pie`,  `histogram-horizontal`, `histogram-vertical`, `violin`.
-5. `config`: the config for the plot
+4. `plot_type`: The type of plot to be displayed i.e `bar`, `pie`, `scatter`, `histogram`, `violin`, and `heatmap`
+6. `config`: the config for the plot
     1. `bargap`: Distance between the bins in the histogram - Only valid for plot type as histogram
     2. `showlegend`: To show the legend or not. DEPRECATED, define in `plotly.layout`
     3. `height`: Height of the plot - DEPRECATED, define in `plotly.layout.height` instead
@@ -60,12 +60,12 @@ Each object in the `plots` array can have the following parameters:
         3. `xaxis.default_all_studies_group`: the `column_name` value from `xaxis.group_keys` array to use as a default when "all studies" is selected on page load
             - takes precedence over `default` property from the list of `group_keys` ONLY when "all studies" is selected
         4. `yaxis.group_key`: similar to x-axis, this matches the `column_name` returned with the data for measuring data on the yaxis
-6. `plotly.config`: the config is ignored if plotly.config is provided. For available options, see the documentation [here](https://plotly.com/javascript/configuration-options/)
+7. `plotly.config`: the config is ignored if plotly.config is provided. For available options, see the documentation [here](https://plotly.com/javascript/configuration-options/)
     1. avoid using `plotly_config`, this property will be DEPRECATED in the future
-7. `plotly.layout`: the `layout` object passed directly to plotly when provided. For available options, see the documentation [here](https://plotly.com/javascript/configuration-options/)
-8. `gene_uri_pattern`: For violin plot only. The url from which the gene data is fetched after applying handlebars templating. This parameter is required to fetch the data for the gene selector and initialize the plot.
-9. `study_uri_pattern`: For violin plot only. The url from which the study data is fetched after applying handlebars templating. This parameter is required to fetch the data for the study selector and display and to initialize the plot.
-10. `traces`: Contains the information about each dataset and how to map that data to plotly.
+8. `plotly.layout`: the `layout` object passed directly to plotly when provided. For available options, see the documentation [here](https://plotly.com/javascript/configuration-options/)
+9. `gene_uri_pattern`: For violin plot only. The url from which the gene data is fetched after applying handlebars templating. This parameter is required to fetch the data for the gene selector and initialize the plot.
+10. `study_uri_pattern`: For violin plot only. The url from which the study data is fetched after applying handlebars templating. This parameter is required to fetch the data for the study selector and display and to initialize the plot.
+11. `traces`: Contains the information about each dataset and how to map that data to plotly.
     1. Properties available to most plot types:
         1. `url_pattern`: The url from which the data has to be fetched
         2. `uri`: The url from which the data has to be fetched. DEPRECATED, use `traces.url_pattern` instead
@@ -88,8 +88,8 @@ Each object in the `plots` array can have the following parameters:
         3. `data_col`: The column name for the values
         4. `legend_col`: An array of column name for the legend to be shown for the respective values
         5. `show_percentage`: To show the percentage or not on slices
-11. `plotlyButtonsToRemove`: The button to be removed shown by plotly by defualt. DEPRECATED, use `plotly.config.modeBarButtonsToRemove` instead
-12. `user_controls`: Contains information about each user control (data for the control, default value, etc) 
+12. `plotlyButtonsToRemove`: The button to be removed shown by plotly by defualt. DEPRECATED, use `plotly.config.modeBarButtonsToRemove` instead
+13. `user_controls`: Contains information about each user control (data for the control, default value, etc) 
     1. `uid`: Key for referencing this user control; in other configuration properties
     2. `label`: Name to display next to this user control
     3. `type`:  The type of control to display, more info below about control types (NOTE: For now, only type `dropdown` is supported)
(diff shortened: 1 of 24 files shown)


src/hooks/selector.ts
```diff
@@ -0,0 +1,97 @@
...
+
+/**
+ * It iterates through its selectorConfig for multiple selectors. 
+ * For each configuration, it extracts values for the selector and sets them in $control_values object within configData.templateParams under the selector's uid
+ * and then returns the modified configData.templateParams.
+ * @param configData Selector configuration, template params and setDataOptions state method
+ * @returns modified configData.templateParams
+ */
+const changeSelectorData = (configData: SelectorsGridProps) => {
+    configData.selectorConfig.map((currentConfig: SelectorConfig)=>{
```
jrchudy, line 57: Throws an error if one plot has a `selectorConfig` defined and the other does not. There should be a check done here or before calling this function in the userEffect

jrchudy:
The variable names and types in the code should be general for the objects in the `user_controls` array. Please change names so that they aren't "selector specific" when they don't need to be. (for example, `SelectorConfig` => `ControlConfig`)

src/components/plot/selectors-grid.tsx
```diff
@@ -0,0 +1,105 @@
...
+    const uid:string[]=[];
+    const valueKey:string[]=[];
+    const selectorValue: Option[]=[];
+    //Collect uid's and valueKey's for all selectors
+    selectorData.selectorConfig?.map((currentConfig,index)=>{
+        const currUid=currentConfig?.uid;
+        const currValueKey=currentConfig?.request_info.value_key;
+        uid.push(currUid);
+        valueKey.push(currValueKey);
+        const selectedOption=selectorOptions[index].find((option: Option) =>
```
jrchudy, line 62: Is this looking through the existing data to initialize a selector? 

I think this makes sense since we should be reading query parameters and setting up `$control_values` if a query parameter is found.

src/hooks/chart.ts
```diff
@@ -269,11 +279,26 @@ export const useChartData = (plot: Plot) => {
     isFetchSelected,
     setIsFetchSelected,
   } = useChartSelectGrid({
+    plot,
     templateParams,
     setModalProps,
     setIsModalOpen,
   });
 
+  useEffect(() => {
```
jrchudy, line 288: Add a comment explaining when this will trigger since it is updating the controls. 

Should this be updated each time `templateParams` is changed? If we have controls that rely on each other, then maybe this makes sense. But I think we should be initializing the selectors once instead of on a change. Does this also mean the selector data is updated when the "plot" changes? (when any of the layout, config, or data changes...?)

src/hooks/selector.ts
```diff
@@ -0,0 +1,97 @@
...
+        const uid = currentConfig?.uid;
+        const valueKey = currentConfig?.request_info?.value_key;
+        const defaultValue = currentConfig?.request_info?.default_value;
+        configData.templateParams.$control_values={
+            ...configData.templateParams.$control_values,
+            [uid]: {
+                values: {},
+            },
+        };
+        if (paramKey) {
```
jrchudy, line 68: This suggests this function is for initializing the selector data and handling it on an update. These should be separate functions since we don't want to parse query parameters again and change values that might have been previously set.

src/models/plot.ts
```diff
@@ -108,10 +134,66 @@ export type TraceConfig = {
...
+  lg: [number, number]; 
+  md: [number, number]; 
+  sm: [number, number]; 
+  xs: [number, number];
+}
+
+/**
+ * Grid Layout config
+ */
+export type GridLayoutConfig = {
```
jrchudy, line 166: Does `ReactGridLayout` have types available in the library we are including?

src/utils/string.ts
```diff
@@ -278,4 +278,34 @@ export const isDataJSON = (data: any) => {
...
+    (_, char) => char.toUpperCase()
+  );
+}
+
+/**
+ * Converts snake_case keys of an object to camel case
+ * @param configObject an object with snake case keys
+ * @returns object with the camel case keys
+ */
+export const gridLayoutConfigMap = (configObject: any) => {
```
jrchudy, line 300: Is this function specific to `gridLayoutConfig` or is it a general solution to converting from one case to another?

Looks like this is a general function for converting keys in an object from snake case to camel case so the function should be named properly so others can use it too.

jrchudy (changes requested):
Other than the inline comments that I left, I noticed a few UX issues that should be addressed:
 - [ ] there is extra whitespace in between the local react grid layout and the plot. 
   - The plot should always be right after the react grid layout
 - [ ] whitespace between the navbar and controls should be reduced
 - [ ] clean up indentation and spacing in the code
 - [ ] add comments for functions in places that the functionality isn't clear
   - every function should have a comment explaining what the expected parameters are and what the function should be doing
   - make sure to include any special cases that the function handles like I asked about for the `changeSelectorData` function

src/hooks/chart.ts
```diff
@@ -131,11 +132,17 @@ export type PlotTemplateParams = {
     [paramKey: string]: any;
   };
   /**
-   * Parameters for URL
+   * Parameters for URL //Deprecated. Use control_values instead
```
jrchudy, line 135: This actually won't be deprecated anymore. We are going to support `url_parameters` in the case that one is defined that is NOT used as part of a control on the page.

For instance, a query parameter is passed to the app that is directly used in the query and can't be changed by the app.

All query parameters will have their `value` stored in `$url_parameters.<key> = <value>` on app initialization. For example, we have `/deriva-webapps/plot/?config=barplot&consortium=ALL` which would have the following `$url_parameters`:
```
$url_parameters: {
  config: 'barplot',
  consortium: 'ALL'
}
```

src/utils/string.ts
KenyShah24: Yes it's general. I'll fix the name

src/models/plot.ts
KenyShah24: Yes it does. I can replace that

src/components/plot/selectors-grid.tsx
KenyShah24: Yeah whatever is the value stored in the $control_values for the given selector should be selected value for that one.

src/hooks/chart.ts
KenyShah24: Got it! I have made changes.

src/hooks/selector.ts
KenyShah24: Yeah but it's just initializing the control values with default ones so I think the function name should be changed and also should be called once. I will update that

src/hooks/chart.ts
KenyShah24: Yes it should be initialized once only.

src/hooks/chart.ts
KenyShah24: Regarding the comment `whitespace between the navbar and controls should be reduced` : This was due to the container_padding I've commented out the parameter from sample_config so that you can test with that

src/hooks/chart.ts
jrchudy: Please add a comment to this useEffect then stating this should trigger only once as setup

docs/user-docs/plot-app.md
```diff
@@ -89,6 +89,40 @@ Each object in the `plots` array can have the following parameters:
         4. `legend_col`: An array of column name for the legend to be shown for the respective values
         5. `show_percentage`: To show the percentage or not on slices
 12. `plotlyButtonsToRemove`: The button to be removed shown by plotly by defualt. DEPRECATED, use `plotly.config.modeBarButtonsToRemove` instead
+13. `user_controls`: Contains information about each user control (data for the control, default value, etc)
```
jrchudy, line 92: Thank you for adding all of this from the issue!

src/components/plot/selectors-grid.tsx
```diff
@@ -0,0 +1,104 @@
...
+        const valueKey = userControlConfig?.request_info?.value_key;
+        templateParams.$control_values[uid].values = {
+            [valueKey]: option.value, //else use default value
+        };
+        return option;
+    }
+};
+
+
+const SelectorsGrid = ({ userControlData, selectorOptions, setSelectorOptionChanged, width }: UserControlsGridProps): JSX.Element => {
```
jrchudy, line 52: This should be change to `UserControlsGrid` and the file should be renamed to `user-controls-grid.tsx`. Make sure to update import statements

src/hooks/chart.ts
```diff
@@ -269,11 +279,26 @@ export const useChartData = (plot: Plot) => {
     isFetchSelected,
     setIsFetchSelected,
   } = useChartSelectGrid({
```
jrchudy, line 281: Can you change this hook name as well? We want to rename it to `useChartControlsGrid`

src/hooks/selector.ts
```diff
@@ -0,0 +1,97 @@
...
+}
+
+/**
+ * It iterates through its selectorConfig for multiple selectors. 
+ * For each configuration, it extracts values for the selector and sets them in $control_values object within configData.templateParams under the selector's uid
+ * and then returns the modified configData.templateParams.
+ * @param configData Selector configuration, template params and setDataOptions state method
+ * @returns modified configData.templateParams
+ */
+const setSelectorData = (configData: UserControlGridProps) => {
```
jrchudy, line 56: Change this to `setControlData`.

jrchudy (changes requested):
A few more changes to take care of still. We discussed the following 2 issues in slack:
 - [ ] change `flex: 1` to `flex: 0 1 0%` for the element with `class="selectors-grid"` to fix spacing between grid and plot
 - [ ] remove `padding: 20px` and `width: 80%` from the element with `class="chart"` to fix spacing above the controls
 - [ ] change a couple more variable names

PR #188, opened 2023-11-08
Matrix border changes

RFSH:
After looking at the existing deployed mousematrix facebase, we noticed new border changes and decided to make similar changes to the React version.

While making these changes, I also refactored the `grid-button.tsx` to use font-awesome instead of SVGs.


RFSH:
@JunmengXu Please make sure to update your branch with the latest master changes to avoid any merge conflicts, and making sure you're working on the latest layout.

JunmengXu:
Got it, will do. Thank you

PR #190, opened 2023-11-17
Plot App: Global layout

KenyShah24:
This feature enables users to display plots and controls in a grid format. Global controls declared universally can manipulate data across all associated plots.

Within each plot, there's an additional grid layout. This layout accommodates user controls specific to that particular plot.

Users have the flexibility to customize grid properties such as columns, margin, container padding, and breakpoints for ResponsiveReactGrid. If not specified, default values will be used. Similarly, users can configure the layout object, responsible for organizing grid items. If not customized, the default layout will stack controls and plots vertically.

For more information : https://github.com/informatics-isi-edu/deriva-webapps/issues/161 (resolve)


src/hooks/control.ts
```diff
@@ -84,33 +81,46 @@ const getDataOptions = async (userControlGridObject: UserControlGridProps) => {
...
-      }
-    }
-    if (defaultValue) {
-      configData.templateParams.$control_values[uid].values = {
-        [valueKey]: defaultValue //else use default value
-      };
-    }
-  })
-  return configData.templateParams;
+export const setControlData = (configData: UserControlConfig[], setTemplateParams: any) => {
```
jrchudy, line 84: I'm unsure why this hook has to be separate from the `useControl` hook. 

We should be collecting the data for the useControl hook before we initialize it instead of initializing it before data is ready and then updating the data after.

src/hooks/control.ts
```diff
@@ -84,33 +81,46 @@ const getDataOptions = async (userControlGridObject: UserControlGridProps) => {
...
+    configData?.map((currentConfig: UserControlConfig) => {
+        const paramKey = currentConfig?.url_param_key;
+        const uid = currentConfig?.uid;
+        const valueKey = currentConfig?.request_info?.value_key;
+        const defaultValue = currentConfig?.request_info?.default_value;
+        if (paramKey) {
+            const paramValue = getQueryParam(windowRef.location.href, paramKey);
+            if (paramValue) {
+            setTemplateParams((prevTemplateParams: any)=>{
```
jrchudy, line 93: This should use `setTemplateParams` from the provider, not a prop passed into this control. We should be able to call `useTemplateParams` in this hook.

src/components/plot/plot-control-grid.tsx
```diff
@@ -0,0 +1,169 @@
...
+const PlotControlGrid = ({
+  config,
+}: PlotControlGridProps): JSX.Element => {
+
+  const [layout, setLayout] = useState<Layouts>({});
+  const [gridProps, setGridProps] = useState({});
+  const [dataOptions, setDataOptions] = useState<any>(null);
+  const [userControlsExists, setUserControlExists] = useState<boolean>(false);
+  const [userControlsReady, setUserControlReady] = useState<boolean>(false);
+  const [initialParams, setInitialParams] = useState<any>({
```
jrchudy, line 37: Shouldn't this be using `templateParams` from the provider?

I think each place in this file where `setInitialParams` is called should be using `setTemplateParams` too. This component shouldn't be keeping track of this state when we are trying to have the state tracked in the provider.

src/components/plot/plot-control-grid.tsx
```diff
@@ -0,0 +1,169 @@
...
+            //i defines the item on which the given layout will be applied
+            i: item?.source_uid,
+            ...item,
+          })))
+        ));
+        setLayout(Object.fromEntries(Object.entries(config.layout).map(([key]: any, index) => [key, mappedLayoutValues[index]])));
+      }
+      //Otherwise set the default layout to display controls and plots 
+      else {
+        const defaultPlotUid = config.plots.map((plot) => {
```
jrchudy, line 65: If there is no `uid` defined in a `plots[x]` configuration, we should use a default of `plot_type` and the index in `plots`. So something like `bar_0` or `violin_0` for some of our existing configurations.

src/components/plot/plot-control-grid.tsx
```diff
@@ -0,0 +1,169 @@
...
+          return plot.uid;
+        });
+        const defaultControlUid = config.user_controls?.map((control) => {
```
jrchudy, line 68: We want a similar default `uid` for controls that we defined for plots. Lets use the control `type`, whether it's `global` or `local`, and it's index in the `user_controls` array. So something like `global_dropdown_0` or `local_dropdown_0`.

src/components/plot/template-params.tsx
```diff
@@ -0,0 +1,47 @@
...
+type TemplateParamsProps = {
+  children: ReactNode;
+}
+
+type ITemplateContext = {templateParams: PlotTemplateParams, setTemplateParams: React.Dispatch<React.SetStateAction<PlotTemplateParams>>,
+  selectorOptionChanged: boolean, setSelectorOptionChanged: React.Dispatch<React.SetStateAction<boolean>> };
+
+// Define the context with the correct initial value
+
+export const TemplateParamsContext = createContext<ITemplateContext>({templateParams: {
```
jrchudy, line 13: Move this to a file in `hooks` similar to what was done in chaise for other context/providers.

src/components/plot/user-control.tsx
```diff
@@ -0,0 +1,72 @@
...
+};
+
+
+const UserControl = ({ controlConfig, controlOptions }: UserControlsGridProps): JSX.Element => {
+    const {templateParams, setTemplateParams, setSelectorOptionChanged} = useContext(TemplateParamsContext);
+    const uid: string=controlConfig.uid;
+    const valueKey: string=controlConfig.request_info.value_key;
+    const selectedOption = controlOptions.find((option: Option) =>
+            option.value === templateParams?.$control_values[uid]?.values[valueKey]);
+    return (
```
jrchudy, line 57: This should be checking for `controlConfig.type` before choosing which user control to display.

src/hooks/chart.ts
```diff
@@ -230,21 +238,14 @@ export const useChartData = (plot: Plot) => {
...
-      },
-      $control_values: {},
-      noData: false, // TODO: remove hack when empty selectedRows are fixed
-    }),
-    []
-  );
+  const {templateParams,setTemplateParams, selectorOptionChanged, setSelectorOptionChanged} = useContext(TemplateParamsContext);
+
+  // To set the dataoptions for the dropdown controllers
+  useControl({
```
jrchudy, line 244: This is being called in `plot-control-grid` too. I thought we discussed that the `useControl` in `plot-control-grid` was setting up global and local controls?

If that's the case, why is it repeated in here? Are the "local" controls actually being initialized here?

src/hooks/chart.ts
```diff
@@ -370,7 +365,7 @@ export const useChartData = (plot: Plot) => {
         if (trace.url_pattern || trace.queryPattern) {
           //To avoid the lint error, passing empty string if both are not present'
           const pattern = trace.url_pattern || trace.queryPattern || '';
-          const { uri, headers } = getPatternUri(pattern, templateParams);
+          const { uri, headers } = getPatternUri(pattern, useInitialParams ? initialParams: templateParams);
```
jrchudy, line 368: Why does this need to be done?

Since we are initializing everything in `plot-control-grid`, templateParams from the provider should be available. There shouldn't be a prop passed to `useChartData` hook for the "initial params". Instead `useChartData` hook should be setup only when the `templateParams` have been initialized in the provider.

src/models/plot.ts
```diff
@@ -205,3 +220,13 @@ export const screenWidthThreshold = 1000;
 
 //Valid file types for url_pattern
 export const validFileTypes = ['csv','json'];
+
+//Default grid layout object to supply to the ResponsiveGridLayout if no grid configuration is provided in the config file
+export const defaultGridProps = {
+  auto_size: true,
+  breakpoints: { lg: 1200, md: 996, sm: 768, xs: 480 },
+  cols: { lg: 12, md: 10, sm: 6, xs: 4 },
+  margin: { lg: [5, 5], md: [10, 10], sm: [9, 9], xs: [5, 5] },
```
jrchudy, line 229: Why do we have different default margins on `md` and `sm`?

jrchudy (changes requested):
- [ ] move template params from components to a folder called `providers` (same folder structure that we have in chaise)
   - add a file to `hooks` folder for creating the context
   - if there's a reason this has to be done the way you did it, we can keep it how it is. Otherwise, we want this to be done the same way everywhere like we did in chaise.
   - this change will allow you to name the "hook" associated with using this context
   - instead of calling `useContext` when you want to use the template params provider, it would be something like `useTemplateParams` instead (or `usePlotApp` which is the name I gave it in my branch)
 - [ ] check how `templateParams`, `initialParams`, and `setTemplateParams` are being used
   - we should be initializing the templateParams in `plot-control-grid` before showing any of the controls or the plots
   - once that's done every sub component and hook should be using the `templateParams` and `setTemplateParams` from the provider
 - [ ] `type` is a required property for `user_control` objects. Please add that and make sure the code isn't making assumptions about what control to display
 
While doing UI testing, for configuration `gudmap-todate-bar-global` I noticed this bug:
image
 
 - [ ] the top "local" dropdown in the screenshot flows "behind" the other dropdown

jrchudy:
Closing this Pull request since this branch was merged into `stage-for-master` which has since been merged into master.

PR #191, opened 2023-11-21
Matrix App: Legend

JunmengXu:
The latest commit focuses on the following issues:

### Legend improvements
- Showing ellipsis or multiple lines for legend labels: Use webkit-line-clamp in CSS to set max lines showing to solve the issue.
  - Modified `src/models/matrix-config.ts` to merge all attributes of Legend to `LegendStyles` and add one more attribute for lineClamp.
  - Modified `src/pages/matrix.tsx` to process default number and pass lineClamp attribute to `Legend` component.
  - Modified `src/components/matrix/legend.tsx` to receive lineClamp attribute and set it as style of `split-text`.
- Make sure the height logic of the grid and legend work properly: Adjust the height of Legend component dynamically.
  - Modified `src/pages/matrix.tsx` to add logic to calculate the height of Legend component to enable the scrollbar of it always shows within the viewport. This is similar to what we did for `gridHeight`.
- Make sure the legend is always horizontally center aligned: Adjust the width of Legend component dynamically.
  - Modified `src/pages/matrix.tsx` to add logic to calculate the width of Legend component with comparison among its children elements and the sum of gridWidth + rowHeaderWidth.
  - Modified `src/assets/scss/_matrix.scss` to set text rotate origin as top left, so that improve the UI clarity and calculation accuracy.
  - Modified `src/components/matrix/legend.tsx` to use `flex` display and `flex-start` alignItems to align bars and texts to avoid aliasing. Modified the style of `legend-links-container` to `marginLeft: barWidth / 2` to align each text as the center of each bar.
- Don't show the hr scrollbar if the legend doesn't need a scroll.
  - Modified `src/assets/scss/_matrix.scss` to add `overflow-x: auto` to the `legend-container` class.

### DIFF
--- a/config/matrix-config-sample.js
+++ b/config/matrix-config-sample.js
@@ -1,4 +1,12 @@
 /**
+ * Please refer to the following for document for more information:
+ * https://github.com/informatics-isi-edu/deriva-webapps/blob/master/docs/user-docs/matrix-app.md#overal-structure
+ *
+ * You can also find the TypeScript definitions here:
+ * https://github.com/informatics-isi-edu/deriva-webapps/blob/master/src/models/matrix-config.ts
+ */
+
+/**
  * The base URI for the data API endpoint
  */
 var baseUri =
@@ -11,17 +19,9 @@ var matrixConfigs = {
      */
     xURL: baseUri + '/sort_key:=X:sort_key,id:=X:id,title:=X:name@sort(sort_key)',
     /**
-     * API for the tree data of the x axis (must return child_id and parent_id. other columns will be ignored.)
-     */
-    xTreeURL: baseUri + 'x_values_tree.json',
-    /**
      * API for the y axis data (must return id and title. other projected columns will be ignored.)
      */
     yURL: baseUri + '/id:=Y:id,title:=Y:name@sort(title)',
-    /**
-     * API for the tree data of the y axis (must return child_id and parent_id. other columns will be ignored.)
-     */
-    yTreeURL: baseUri + 'y_values_tree.json',
     /**
      * API for the z axis data (color axis) (must return id and title. other projected columns will be ignored.)
      */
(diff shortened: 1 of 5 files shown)


RFSH (changes requested):
We had a conversation about the changes in this PR, and the following is the summary of it:

- The width and line clamp changes look good and we should push them.
- We should revert the height changes in this PR. In these changes, we're assuming that the navbar and top portion of the page have a fixed height, which is usually not the case. That being said, the existing code is also not doing the right thing. So we should try to come up with a better solution in another PR.

JunmengXu:
The latest commit focuses on the following issues:

- Revert the height changes logic of Grid and Legend in this PR, to original version. So the second improvement point in the previous comment was removed.

  - Modified `src/pages/matrix.tsx` to remove calculation logic of the height of Legend component, and revert code to original version.

JunmengXu:
In terms of `matrix-config.js` configuration, the new format for the legend should be as below:

```
legend: {
    /**
    * Height of the legend
    */
    height: 100,
    /**
     * Width of the legend bar
     */
    barWidth: 55,
    /**
     * Height of the legend bar
     */
    barHeight: 15,
    /**
     * Max number of lines showing the legend content
     */
    lineClamp: 1,
}
```

RFSH (approved):

RFSH:
@JunmengXu Everything looks good to me. Please also update the [`config/matrix-config-sample.js`](https://github.com/informatics-isi-edu/deriva-webapps/blob/master/config/matrix-config-sample.js) file to reflect the changes that you did to legend styles.

JunmengXu:
The latest commit focuses on the following issues:

- Updated `config/matrix-config-sample.js` file to add configurations for legend, as previous comment shown.
- Updated `config/matrix-config-sample.js` file to add configurations for matrix-treeview, specifically, `xTreeURL` and `yTreeURL` for hierarchical data, wrap up properties of the row headers and column header to `rowHeader` and `columnHeader`.

Issue #199, opened 2024-02-07
Change layout to layouts and move to grid_layout_config


jrchudy:
We should change `plots[0].layout` to be `plots[0].grid_layout_config.layouts` to be consistent with the react-grid-layout language. This way it’s in the same “object” that react grid layout has in its configuration language for responsive plots (and the name is the same, layouts is plural). [Link](https://github.com/react-grid-layout/react-grid-layout?tab=readme-ov-file#responsive-grid-layout-props) to react grid layout documentation.

This will also hopefully prevent confusion with what `plots[0].layout` maybe meaning something similar to `plotly.layout`.

This should be done for both the global and local react grid layouts in our code.

PR #200, opened 2024-02-07
Plot App: layouts parameter added to grid_layout_config for global and local selectors

KenyShah24:
Changed plots[0].layout to be plots[0].grid_layout_config.layouts to be consistent with the react-grid-layout language. This will prevent confusion with what plots[0].layout maybe meaning something similar to plotly.layout.

Resolves: #199


jrchudy (approved):
Changes look good. Thank you for updating the sample config document.

Tested these changes and everything is working as expected

Issue #201, opened 2024-02-13
React grid layout configuration validation


jrchudy:
We want to validate the new properties that are being added to the plot configuration. `grid_layout_config` and each control in `user_controls` should be checked for:
 - if those properties are required
   - if a property has defaults we can fall back to, then it isn't required and we should use the default when not defined
 - the property's expected type whether it's a string, array, object etc
   - if not defined properly, maybe we want to show a warning alert
   - fall back to default value
   - if property is required, alert should maybe be an error instead of a warning

The following is from a https://github.com/informatics-isi-edu/deriva-webapps/issues/161#issuecomment-1791397832

We want to list the properties in the `grid_layout_config`, each control in `user_controls`, and each layout in `layout` that are required to be defined as well as checking the type of the set value. We should be validating this list in the code and throwing proper errors when those properties are not defined. 

List of properties to validate:
 - `grid_layout_config`:
   - `cols` - might not need to be validated since we might have a default to use? 12 cols as default maybe?
     - fails validation for this control if not defined
       - **what does a failed validation mean?**
   - _please fill in more..._
 - `user_control`:
   - `uid` - required or can't map control to `$control_values` used for templating or map to an object in `layout`
     - fails validation for this control if not defined
   - `type` - required unless we want to assume a default is dropdown
     - we should probably enforce this as a requirement since other properties being defined depend on this being set
     - fails validation for this control if not defined
       - **what does a failed validation mean?**
         - don't add to set of user controls and don't load the control in the UI
         - put a warning in browser as alert OR warning in browser console
           - warning icon doesn't make sense since this control won't be displayed
           - error popup doesn't make sense since this shouldn't break the app
   - _please fill in more..._
 - `layout`:
   - `source_uid` - required or we don't know what control/plot to show
     - fails validation if not defined
       - **what does a failed validation mean?**
         - ignore this layout entry
         - put a warning in browser as alert OR warning in browser console
   - _please fill in more..._

Types of errors might include:
 - error modal popup
 - alert warning
 - browser console output
 - warning icon in the component that can't be loaded or have data fetched for
   - how we do this for facet panel:
      <img width="351" alt="image" src="https://github.com/informatics-isi-edu/deriva-webapps/assets/2932600/382f8f6f-ba32-46c6-9748-4f0fef51b9b6">
   - how chrome does it for developer tools:
[...]

PR #204, opened 2024-03-06
Plot grid layout validation

KenyShah24:
Resolves: #201 
Validation issues handled:

1. grid_layout_config -> cols: Ignored/Handled and app will work fine.
2. user_control -> uid : Blocking error will be displayed for local and global controls
3. user_control -> type: Control won't be shown and an alert will be shown with error.
4. user_control -> label: No label will be shown and no error will be shown
5. user_control .request_info.data/url_pattern: Control won't be shown & an alert will be shown with type warning
6. layout -> source_uid: An alert will be shown with type error and the control will be shown without any layout parameters applied to it.
7. layout -> w/h/x/y: Control/plot won't be shown and an alert will be shown with the type warning
8. Plots/Control with same uid: An alert will be shown with type error and the latest component with the uid will be shown after overriding the earlier one's with same uid.

### DIFF
--- a/src/components/plot/plot-control-grid.tsx
+++ b/src/components/plot/plot-control-grid.tsx
@@ -65,11 +65,6 @@ const PlotControlGrid = ({
   useEffect(() => {
     if (!globalControlsInitialized) return;
 
-    if (userControlsExists && templateParams?.$control_values) {
-      // userControlConfig should exist if userControlsExists === true
-      if (globalUserControlData.userControlConfig) setUserControls(globalUserControlData.userControlConfig);
-    }
-
     if (userControlsExists && globalUserControlData?.userControlConfig?.length > 0) {
       //Validate global controls for uid and type. Display error if any control doesn't have an uid or type
       validateUID(globalUserControlData?.userControlConfig, alertFunctions);
--- a/src/components/plot/plot-control-grid.tsx
+++ b/src/components/plot/plot-control-grid.tsx
@@ -12,16 +12,15 @@ import usePlot from '@isrd-isi-edu/deriva-webapps/src/hooks/plot';
 import { useEffect, useRef, useState } from 'react';
 
 // models
-import { DataConfig, LayoutConfig, Plot, UserControlConfig, defaultGridProps, globalGridMargin } from '@isrd-isi-edu/deriva-webapps/src/models/plot';
-import { Layout, Layouts, Responsive, WidthProvider } from 'react-grid-layout';
+import { DataConfig, Plot, UserControlConfig, defaultGridProps, globalGridMargin } from '@isrd-isi-edu/deriva-webapps/src/models/plot';
+import { Layouts, Responsive, WidthProvider } from 'react-grid-layout';
 
 // provider
 import PlotlyChartProvider from '@isrd-isi-edu/deriva-webapps/src/providers/plotly-chart';
 
 // utils
 import useAlert from '@isrd-isi-edu/chaise/src/hooks/alerts';
-import { ChaiseError } from '@isrd-isi-edu/chaise/src/models/errors';
-import { ChaiseAlertType } from '@isrd-isi-edu/chaise/src/providers/alerts';
+import { validateControlData, validateDuplicateControlUID, validateLayout, validateUID } from '@isrd-isi-edu/deriva-webapps/src/utils/plot-utils';
 import { convertKeysSnakeToCamel, validateGridProps } from '@isrd-isi-edu/deriva-webapps/src/utils/string';
 
 export type PlotControlGridProps = {
@@ -40,7 +39,7 @@ const PlotControlGrid = ({
   const [layout, setLayout] = useState<Layouts>({});
   const [gridProps, setGridProps] = useState({});
   const [userControlsExists, setUserControlExists] = useState<boolean>(false);
-  const [userControlsReady, setUserControlReady] = useState<boolean>(false);
+  const [userControlsReady, setUserControlsReady] = useState<boolean>(false);
   const [userControls, setUserControls] = useState<UserControlConfig[]>([]);
   const [validatedPlots, setValidatedPlots] = useState<Plot[]>(config.plots);
   const alertFunctions = useAlert();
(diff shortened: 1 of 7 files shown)


src/components/plot/plot-control-grid.tsx
```diff
@@ -126,17 +261,15 @@ const PlotControlGrid = ({
         })]
       ))
     }
-
     setLayout(tempLayout);
-  }, [globalControlsInitialized]);
+  },[userControls,validatedPlots]);
```
jrchudy, line 265: instead of `userControls`, this should trigger on `userControlsReady`.

src/components/plot/plot-control-grid.tsx
```diff
@@ -38,7 +41,10 @@ const PlotControlGrid = ({
   const [gridProps, setGridProps] = useState({});
   const [userControlsExists, setUserControlExists] = useState<boolean>(false);
   const [userControlsReady, setUserControlReady] = useState<boolean>(false);
```
jrchudy, line 43: I noticed an inconsistency in this state variable. we have `userControlsReady` and `setUserControlReady` (the `set` is not plural).

```
setUserControlReady --> setUserControlsReady
```

Can you change this to be consistent and update the places in this file where the set function is called?

src/components/plot/user-controls-grid.tsx
```diff
@@ -33,27 +37,123 @@ const UserControlsGrid = ({
...
+  /**
+  * It should be called once to validate the configuration data for the user controls
+  */
   useEffect(() => {
-    if (userControlData?.gridConfig?.layouts && Object.values(userControlData.gridConfig?.layouts).length > 0) {
+    if (setupStarted.current) return;
+    setupStarted.current = true;
+    if (localControlData?.length > 0) {
+
+      //Validate local controls for uid and type. Display error if any control doesn't have an uid or type
```
jrchudy, line 58: The following is the same check done in `plot-control-grid`. Could you factor this check out into a separate function in a utils file (`plot-utils.ts` should be fine for now).

src/components/plot/user-controls-grid.tsx
```diff
@@ -33,27 +37,123 @@ const UserControlsGrid = ({
...
+        if (!control?.type) {
+          invalidControlUid.push(control.uid);
+        }
+      });
+      if (invalidControlUid.length > 0) {
+        alertFunctions.addAlert(`Unable to display the local control with Uid(s) '${invalidControlUid.join(', ')}' as the 'type' was not found for the control`, ChaiseAlertType.ERROR);
+      }
+    }
+
+    //Validate local controls with same uid. Display error in case of same uid(s)
```
jrchudy, line 73: Same comment as above, this validation is also duplicated from `plot-control-grid`. This should be a 2nd function in `plot-utils.ts`.

src/components/plot/user-controls-grid.tsx
```diff
@@ -33,27 +37,123 @@ const UserControlsGrid = ({
...
+      userControlData?.userControlConfig?.map((control: UserControlConfig) => {
+        if (!control?.uid) {
+          throw new ChaiseError('User Control Error', `Missing UID for the specified global user control  ${control?.label ? `labelled as "${control?.label}"` : ''}`);
+        }
+        if (!control?.type) {
+          invalidControlUid.push(control.uid);
+        }
+      });
+      if (invalidControlUid.length > 0) {
+        alertFunctions.addAlert(`Unable to display the local control with Uid(s) '${invalidControlUid.join(', ')}' as the 'type' was not found for the control`, ChaiseAlertType.ERROR);
```
jrchudy, line 69: Change `Uid(s)` to `UID(s)` here

src/components/plot/user-controls-grid.tsx
```diff
@@ -33,27 +37,123 @@ const UserControlsGrid = ({
...
-    if (userControlData?.gridConfig?.layouts && Object.values(userControlData.gridConfig?.layouts).length > 0) {
+    if (setupStarted.current) return;
+    setupStarted.current = true;
+    if (localControlData?.length > 0) {
+
+      //Validate local controls for uid and type. Display error if any control doesn't have an uid or type
+      const invalidControlUid: string[] = [];
+      userControlData?.userControlConfig?.map((control: UserControlConfig) => {
+        if (!control?.uid) {
+          throw new ChaiseError('User Control Error', `Missing UID for the specified global user control  ${control?.label ? `labelled as "${control?.label}"` : ''}`);
```
jrchudy, line 62: Same comment as above about using index if label is also undefined. This should be handled when the function is factored out.

src/components/plot/user-controls-grid.tsx
```diff
@@ -33,27 +37,123 @@ const UserControlsGrid = ({
...
+          dupUid.push(control.uid);
+        }
+      }
+      result[control.uid] = control;
+      return result;
+    }, {});
+    if (dupUid?.length > 0) {
+      dupUid.map((id) => alertFunctions.addAlert(`Multiple local controls with the same UID '${id}' were detected`, ChaiseAlertType.ERROR));
+    }
+    userControlData.userControlConfig = Object.values(controlObject);
```
jrchudy, line 90: We want to avoid modifying the props that are passed to this component. Do we need to update this or is it sufficient to keep the state in this component after validating?

I think it should be sufficient to keep the state in the component since it shouldn't be used outside of it.

src/components/plot/user-controls-grid.tsx
```diff
@@ -33,27 +37,123 @@ const UserControlsGrid = ({
...
+    setLocalControlData(Object.values(controlObject));
```
jrchudy, line 91: Avoid setting this multiple times. This should be done after all validations.

src/components/plot/user-controls-grid.tsx
```diff
@@ -33,27 +37,123 @@ const UserControlsGrid = ({
...
+          (alert) => alert.message.includes(`Unable to display the local control with Uid '${control.uid}'`))) {
+          invalidControlData.push(control.uid);
+        }
+      }
+      return ({ ...control, visible: isControlValid });
+    }).filter((control) => control.visible || control?.visible == undefined);
+    if (invalidControlData?.length > 0) {
+      alertFunctions.addAlert(`Unable to display the local control with Uid(s): ${invalidControlData.join(', ')} because neither the request_info.data nor the request_info.url_pattern were found`, ChaiseAlertType.WARNING);
+    }
+    setLocalControlData(validatedUserControls);
```
jrchudy, line 109: Avoid setting this multiple times. This should be done after all validations.

src/components/plot/user-controls-grid.tsx
```diff
@@ -33,27 +37,123 @@ const UserControlsGrid = ({
...
+      result[control.uid] = control;
+      return result;
+    }, {});
+    if (dupUid?.length > 0) {
+      dupUid.map((id) => alertFunctions.addAlert(`Multiple local controls with the same UID '${id}' were detected`, ChaiseAlertType.ERROR));
+    }
+    userControlData.userControlConfig = Object.values(controlObject);
+    setLocalControlData(Object.values(controlObject));
+
+    // Validate controls for not having neither of the request_info.data nor the request_info.url_pattern and display error
```
jrchudy, line 93: Another place where a function for validating is done in both files. Please factor this out into a 3rd function in `plot-utils`

src/components/plot/user-controls-grid.tsx
```diff
@@ -33,27 +37,123 @@ const UserControlsGrid = ({
...
+    const invalidControlData: string[] = [];
+    const validatedUserControls = userControlData.userControlConfig?.map((control: UserControlConfig) => {
+      let isControlValid = true;
+      if (!((control.request_info?.data && control.request_info?.data?.length > 0) || control.request_info?.url_pattern)) {
+        isControlValid = false;
+        if (!alertFunctions.alerts.some(
+          (alert) => alert.message.includes(`Unable to display the local control with Uid '${control.uid}'`))) {
```
jrchudy, line 100: `Uid` should be `UID` for consistency

src/components/plot/user-controls-grid.tsx
```diff
@@ -33,27 +37,123 @@ const UserControlsGrid = ({
...
+          invalidControlData.push(control.uid);
+        }
+      }
+      return ({ ...control, visible: isControlValid });
+    }).filter((control) => control.visible || control?.visible == undefined);
+    if (invalidControlData?.length > 0) {
+      alertFunctions.addAlert(`Unable to display the local control with Uid(s): ${invalidControlData.join(', ')} because neither the request_info.data nor the request_info.url_pattern were found`, ChaiseAlertType.WARNING);
```
jrchudy, line 107: Same comment about `Uid(s)` being `UID(s)` instead.

src/components/plot/user-controls-grid.tsx
```diff
@@ -33,27 +37,123 @@ const UserControlsGrid = ({
...
+      // Check if the layout parameter has all the necessary properties such as source_uid, h, w, x, and y. If any of these properties are missing, display a warning.
+      tempLayout = Object.fromEntries(Object.entries(tempLayout).map(([layoutBPKey, layoutBPValue]) => {
+        const newValue: Layout[] = (layoutBPValue as any[]).filter((val: any) => {
+          if (!val?.sourceUid) {
+            alertFunctions.addAlert(`Unable to display the local control due to a missing 'source_uid' in layouts parameter for "${layoutBPKey}" breakpoint`, ChaiseAlertType.WARNING);
+            return false;
+          }
+          //Not showing single error in this case as it's specific to a breakpoint
+          if (!(val?.w !== 0 && val?.w !== undefined && val?.h !== 0 && val?.h !== undefined && val?.x !== undefined && val?.y !== undefined)) {
+            alertFunctions.addAlert(`Unable to display the local control with Uid "${val.sourceUid}" due to a one of the required layout parameters missing(w, h, x, y) for "${layoutBPKey}" breakpoint`, ChaiseAlertType.WARNING);
```
jrchudy, line 132: Another place to change `Uid` to `UID`

src/components/plot/user-controls-grid.tsx
```diff
@@ -33,27 +37,123 @@ const UserControlsGrid = ({
...
           ...item,
         })))
       ));
-
-      setLayout(Object.fromEntries(Object.entries(userControlData.gridConfig?.layouts).map(([key]: any, index) => [key, mappedLayoutValues[index]])));
+      let tempLayout;
+      tempLayout = Object.fromEntries(Object.entries(userControlData.gridConfig?.layouts).map(([key]: any, index) => [key, mappedLayoutValues[index]]));
+      const invalidLayoutUid: string[] = [];
+
+      // Check if the layout parameter has all the necessary properties such as source_uid, h, w, x, and y. If any of these properties are missing, display a warning.
```
jrchudy, line 123: This is a 4th function we can factor out to `plot-utils` since we are validating the layout the same way in both places.

src/components/plot/user-controls-grid.tsx
```diff
@@ -33,27 +37,123 @@ const UserControlsGrid = ({
...
+      //Revalidate to filter out controls with invalid layout
+      const revalidatedUserControls = validatedUserControls?.map((control: UserControlConfig) => {
+        let isControlValid = true;
+        if (invalidLayoutUid.includes(control?.uid)) {
+          isControlValid = false;
+          return ({ ...control, visible: isControlValid });
+        }
+        return ({ ...control });
+      }).filter((control) => control.visible || control?.visible == undefined);
+      setLocalControlData(revalidatedUserControls);
```
jrchudy, line 150: Don't call this here and instead we will call it after the `else` case below. So we only set the `localControlData` once in this useEffect. See comment below

src/components/plot/user-controls-grid.tsx
```diff
@@ -65,18 +165,19 @@ const UserControlsGrid = ({
...
               static: true
             }
           })]
         }
       ))
-
       setLayout(tempLayout);
     }
+    //To avoid react-grid-layout throwing unnecessary errors related to layout or UID
+    setValidationComplete(true);
```
jrchudy, line 180: call `setLocalControlData()` with `revalidatedUserControls` OR `validatedUserControls` if they were revalidated.

```
if (revalidatedUserControls && revalidatedUserControls.length !== -1) {
  setLocalControlData(revalidatedUserControls);
} else {
  setLocalControlData(validatedUserControls)
}
```

jrchudy (changes requested):
There are some changes that we want to make based on the comments I left in the code:
 - [x] avoid calling `setUserControls` multiple times in the same `useEffect`
   - the same is true for `setLocalControlData`
 - [x] factor out validation functions for each step. Have a separate function for each validation
   1. validate controls for `uid` and `type`
   2. validate controls for duplicate `uid` values
   3. validate controls for request_info.data or request_info.url_pattern
   4. validate layout
 - [ ] create an enum for the control types 
 - [ ] consistent use of `UID`

jrchudy (changes requested):
A few other changes then this PR will be good to merge:
 - [x] some spacing issues in the validation functions
 - [ ] adjust spelling in some alerts/error messages to include `.` and remove unnecessary whitespace character (` `) and `a`
 - [ ] global error that incorrectly has `local` in the error message
 - [ ] layout validation message displayed 3 times for `plot-control-grid` component

jrchudy (approved):
Changes look good, I will test this once more in the morning and merge these changes.

PR #209, opened 2024-03-27
Plot App: Added app_styles

### LINKED ISSUE #163: Configurable plot app container
The plot app currently resizes to the size of its parent container, but in plot app part of the setup defines "dynamic styles" for the `PlotlyChart` component. This dynamic style is hardcoding the width to `80%` of the plot app container width (browser width or iframe width).

As part of the changes for issue #160 , we are changing this calculation to `95%` instead so plot uses the whitespace better on smaller screens and in iframes (dev.gudmap.org has a bar plot in an iframe for example).

We want to include a property in plot-config for data modelers to specify what this container width should be. Since this is for the "whole" plot app (width on either side of all plots and selectors), it should be at the top level of the plot configuration next to `plots` called `app_width`:
```
'config-name': {
  `app_width`:             percentage width to use for the whole plot app (as percent or decimal?)
  `headTitle`:             ...,
  `top_right_link_text`:   ...,
  `plots`:                 [ {...}, ... ]
}
```

Needs discussion:
 - should `app-width` be a percent value (0-100) or a decimal (0.0-1.0)?
 - should other formats be supported (pixel width)
   - alternative solution: this can be achieved by defining the width of the iframe but doesn't change this value inside the plot app. results may not be as expected

### LINKED ISSUE #176: Max width and height configuration parameters
Currently on rbk/gudmap/atlas we have 2 configurations per consortium (6 total) that are nearly identical (`*-data-summary` and `*-data-summary-responsive`). The only difference is that the height and width of the plot are defined in the non `responsive` configurations. Our code then fixes the height and width of the plot container.

When the height and width are not defined, the plot container instead uses the following height and width:
```
width = 0.95 * containerWidth (app_width proposed in another issue to modify this 0.95)
height = 0.7 * containerHeight
```

In this case, the max width and height seems to be set appropriately for responsive plots. We should allow for these 2 properties to be defined in the `plot-config` document. We will add them at the first level of the document alongside `plots` since these properties will affect the height and width of the container that the plots are contained in (to resize properly to the "viewport" of the plot, whether it's standalone or in an iframe).
```
var plotConfigs = {
  "config-name": {
    headTitle: "...",
    max_app_height:    Max height of the container surrounding the plots
    max_app_width:     Max width of the container surrounding the plots
    plots: [ ... ]
  }
}
```

KenyShah24:
- Max height and width for plot app : Resolves #176 
- Width for plot app : Resolves #163 
- Following 3 different properties will be added to the configuration for plot app

```
'config-name': {
  app_styles: {       // intended to only attach styles to element with `app_container` class
    width:  '',       // percentage width to use for the whole plot app as a decimal from 0.0-1.0
    max_width: '',    // same values allowed as property in CSS
    max_height: '',   // same values allowed as property in CSS
  },
...
}
```


src/components/plot/plot-control-grid.tsx
```diff
@@ -22,13 +22,13 @@ import PlotlyChartProvider from '@isrd-isi-edu/deriva-webapps/src/providers/plot
 import useAlert from '@isrd-isi-edu/chaise/src/hooks/alerts';
 import { validateControlData, validateDuplicateControlUID, validateLayout, validateUID } from '@isrd-isi-edu/deriva-webapps/src/utils/plot-utils';
 import { convertKeysSnakeToCamel, validateGridProps } from '@isrd-isi-edu/deriva-webapps/src/utils/string';
+import { useWindowSize } from '../../hooks/window-size';
```
jrchudy, line 25: Update this import to the direct path instead of relative.

src/components/plot/chart-with-effect.tsx
```diff
@@ -168,7 +170,7 @@ const ChartWithEffect = ({ config }: ChartWithEffectProps): JSX.Element => {
     }
   }
   return (
-    <div className='chart-container'>
+    <div className='chart-container' style={{maxWidth: appStyles?.max_width, maxHeight: appStyles?.max_height}}>
```
jrchudy, line 173: These properties were intended to be for the whole app, not each specific chart in react grid layout. We should have them on the html element with class `plot-page` in `plot-control-grid.tsx`. This way the width we defineon `grid-container` will adjust based on a set `max-width`

jrchudy (changes requested):
Please change where the `max_width` and `max_height` are being applied based on my comment. These 2 properties should be on the `plot-page` html element so that the component with `width` defined on it will resize based on this `max_width`

jrchudy (approved):

PR #220, opened 2024-11-11
Remove vitessce app and update dependencies

RFSH:
Adding Vitessce to our dependencies has significantly increased the installation and build time. Since none of our deployments currently use this app, we decided to remove it for now. So, in this PR, 

- Modified `package.json` to not include Vitesse anymore. 
- Commented the Vitessce app definition from webpack configuration.
- Removed Vitessce from deploy makefile targets.

I also updated the included version of chaise and other dependencies to get rid of the warning/error messages.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -149,10 +149,10 @@ run-webpack: $(BUILD_VERSION)
 
 #exclude <app>-config.js to not override one on deployment
 .PHONY: deploy
-deploy: dont_deploy_in_root print-variables deploy-boolean-search deploy-heatmap deploy-plot deploy-treeview deploy-matrix deploy-vitessce
+deploy: dont_deploy_in_root print-variables deploy-boolean-search deploy-heatmap deploy-plot deploy-treeview deploy-matrix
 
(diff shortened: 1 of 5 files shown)


KenyShah24:
The Plot, Heatmap, and Matrix apps are functioning well with no issues observed related to the dependency update or the removal of the Vitessce app.

Here are the issues I encountered while testing various apps:

1. Plot App: In the violin plot, there is a minor issue where the selector grid text is getting cut off. This issue also exists in the Master branch.
Screenshot 2024-11-11 141129

2. Treeview App: The app is not visible when using the default configuration in both the Treeview and Boolean Search apps. This issue is also present in the Master branch.
image

RFSH:
@KenyShah24 Thank you for double-checking the apps. The problem with treeview under your user directory is the `treeview-config.js` that you're using. If you use the same config that staging.atlas-d2k.org is using (like how I'm doing [here](https://staging.atlas-d2k.org/~ashafaei/deriva-webapps/treeview/)), the app loads properly.

PR #226, opened 2025-05-05
Added x, y and z col pattern feature for plot config

KenyShah24:
Added new `col_pattern` paramaters which has `control_value` templates for x, y and z col values to select the column names dynamically using `user_controls`.
1. `x_col_pattern`: array of user control(`user_controls`) patterns to turn into display values to use for x column name for x values.
2. `y_col_pattern`: array of user control(`user_controls`) patterns to turn into display values to use for y column name for y values.
3. `z_col_pattern`(For heatmap charts): array of user control(`user_controls`) patterns to turn into display values to use as the aggregated value for the `(x, y)` combination

When `col_pattern` and `col` parameters are given, `col_pattern` will be prioritize over static col values.


RFSH (changes requested):
Apart from the minor changes that I asked for, please also modify one of the existing configs to use the pattern feature that you've added in this PR.

RFSH (approved):

